PR #1, opened 2018-06-28
Propose an RFC process based on Rust's but lighter weight

fitzgen:
[RFC rendered](https://github.com/rustwasm/rfcs/blob/001-the-rfc-process/text/001-the-rfc-process.md)

[Template rendered](https://github.com/rustwasm/rfcs/blob/001-the-rfc-process/000-template.md)

### DIFF
--- a/text/001-the-rfc-process.md
+++ b/text/001-the-rfc-process.md
@@ -1,0 +1,220 @@
+- Start Date: 2018-06-28
+- RFC PR: (leave this empty)
+- Tracking Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Adopt a simplified version of the Rust RFC process to the Rust and WebAssembly
+domain working group. The RFC process will provide a single place to decide on
+substantial changes and additions across the ecosystem for all stakeholders.
+
+# Motivation
+[motivation]: #motivation
+
+There are some decisions which have broad impact across the Rust and WebAssembly
+ecosystem, and therefore have many stakeholders who deserve to have a say in the
+decision and provide feedback on proposals and designs. Right now, these
+decisions tend to be made in whatever local repository pull request or issue
+tracker. This makes it difficult for stakeholders to stay on top of these
+decisions, because they need to watch many different places. For a repository
+owner or team, it is also difficult to determine whether the ecosystem is in
+favor of a feature or not.
+
+After adopting this RFC process, stakeholders should have an easier time staying
+on top of substantial changes and features within the ecosystem. Additionally,
+the maintainers of a particular repository within the Rust and WebAssembly
+ecosystem should feel confident that they've solicited feedback from everyone
+involved after going through an RFC, and won't get angry bug reports from users
+who felt that they were not consulted. Everyone should have shared confidence in
+the direction that the ecosystem evolves in.
+
+# Detailed Explanation
+[detailed-explanation]: #detailed-explanation
+
+Right now, governance for repositories within the `rustwasm` organization
+[follow these rules][repo-governance] describing policy for merging pull
+requests:
+
+> Unless otherwise noted, each `rustwasm/*` repository has the following general
+> policies:
+>
+> * All pull requests must be reviewed and approved of by at least one relevant
+>   team member or repository collaborator before merging.
+>
+> * Larger, more nuanced decisions about design, architecture, breaking changes,
+>   trade offs, etc are made by the relevant team and/or repository
+>   collaborators consensus. In other words, decisions on things that aren't
+>   straightforward improvements to or bug fixes for things that already exist
+>   in the project.
+
+This policy categorizes pull requests as either "larger, more nuanced ..."
+changes or not (I will use "substantial" from now on). When a change is not
+substantial, it requires only a single team member approve of it. When a change
+is larger and more substantial, then the relevant team comes to consensus on how
+to proceed.
+
+This RFC intends to further sub-categorize substantial changes into those that
+affect only maintenance of the repository itself, and are therefore only
+substantial *internally* to the maintainers, versus those substantial changes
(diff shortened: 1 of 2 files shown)


fitzgen:
I just thought of another way we could potentially speed up our lighter-weight version of Rust's RFC process: make voting on an RFC happen in parallel with waiting on N days of FCP before merging, rather than waiting on votes and then sequentially starting N days of FCP.

```
----------------------------- time ------------------------------>

                    Rust's RFC Process
                    ==================

    |-------- FCP proposal --------|------- FCP ----------|
    ^                              ^                      ^
    |                              |                      |
    |                              |                      |
Team member                     All team                N days of FCP
proposes FCP                    members sign            have passed, so RFC
                                off                     is merged/closed

                    Lighter-weight RFC Process?
                    ===========================

                                       N days of FCP occur
                                       concurrently with team
                                       voting yay or nay on
                                       the RFC
                                           |
                                           V
    |------------------- FCP --------------|
    |-------- Team Voting ---------|
    ^                              ^
    |                              |
    |                              |
Team member                     All team
proposes vote                   members sign
                                off, but we
                                still have to
                                wait for N
                                days of FCP to
                                complete
```

mgattozzi:
What if a team member says we shouldn't move to FCP then? We'd already be in the middle of it. My understanding was that the team voting initiates FCP

text/001-the-rfc-process.md
```diff
@@ -0,0 +1,220 @@
...
+ecosystem should feel confident that they've solicited feedback from everyone
+involved after going through an RFC, and won't get angry bug reports from users
+who felt that they were not consulted. Everyone should have shared confidence in
+the direction that the ecosystem evolves in.
+
+# Detailed Explanation
+[detailed-explanation]: #detailed-explanation
+
+Right now, governance for repositories within the `rustwasm` organization
+[follow these rules][repo-governance] describing policy for merging pull
```
mgattozzi, line 36: Maybe this should be permalinked to the current commit for posterity sake

text/001-the-rfc-process.md
```diff
@@ -0,0 +1,220 @@
...
+    if it is made prematurely.
+  - For RFCs with lengthy discussion, the motion to FCP should be preceded by a
+    *summary comment* trying to lay out the current state of the discussion and
+    major tradeoffs/points of disagreement.
+  - Before actually entering FCP, *all* members of the team(s) must sign off;
+    this is often the point at which many team members first review the RFC in
+    full depth.
+- The FCP lasts seven calendar days. It is also advertised widely, e.g. in an
+  issue of ["This Week in Rust and WebAssembly" on the Rust and WebAssembly
+  blog](https://rustwasm.github.io/). This way all stakeholders have a chance to
```
mgattozzi, line 139: Maybe we can also bring it up on ThisWeekInRust

text/001-the-rfc-process.md
```diff
@@ -0,0 +1,220 @@
...
+
+- FCP lasts seven calendar days rather than ten. This reflects our desire for a
+  lighter-weight process that moves more quickly than Rust's RFC process.
+- The RFC template is shorter and merges together into single sections what were
+  distinct sections in the Rust RFC template. Again, this reflects our desire
+  for a lighter-weight process where we do not need to go into quite as much
+  painstaking detail as Rust RFCs sometimes do (perhaps excluding *this* RFC).
+
+The phases of RFC development and post-RFC implementation are largely the same
+as the Rust RFC process. I found that the motivations for nearly every phase of
```
mgattozzi: It might be better to remove uses of I in an RFC as it also serves as a reference if accepted. So it should be a bit more impersonal in nature.

text/001-the-rfc-process.md
```diff
@@ -0,0 +1,220 @@
...
+We can also avoid adopting an RFC process, and move more quickly by allowing
+each repository's team or owner to be dictators of their corner of the
+ecosystem. However, this will result in valuable feedback, opinions, and insight
+not getting voiced, and narrow decisions being made.
+
+# Unresolved Questions
+[unresolved]: #unresolved-questions
+
+- Will we use [`@rfcbot`][rfcbot]? I expect that if we can, we should, but this
+  can be decided separately from whether to accept this RFC.
```
mgattozzi: This would be cool

fitzgen:
> What if a team member says we shouldn't move to FCP then? We'd already be in the middle of it. My understanding was that the team voting initiates FCP

I was imagining that it would cancel the FCP and we would go back to development, the same way new concerns raised during FCP can send the RFC back to development. Open to other suggestions as well, or also keeping the serialized voting and then FCP.

fitzgen:
Apparently getting @rfcbot setup is pretty much a PR for https://github.com/anp/rfcbot-rs/blob/master/rfcbot.toml and https://github.com/anp/rfcbot-rs/blob/fd4e44f531e6127ed122921c16387dff5e7ac522/src/config.rs#L7 plus some permissions and webhooks stuff for this GH org.

fitzgen:
During our last WG meeting, @ashleygwilliams suggested adding some sort of "here is the outreach I've done to outside stakeholders" section to the template. I'll add something like that and then probably propose FCP (without the concurrent FCP and signing off discussed above).

fitzgen:
I am proposing we enter Final Comment Period for this RFC that establishes our RFC process! :)

Disposition: **merge**

* [X] @fitzgen
* [x] @ashleygwilliams 
* [x] @alexcrichton

alexcrichton:
👍 Looks great to me!

ashleygwilliams:
very excited for this. thanks for the effort @fitzgen

fitzgen:
Enter the *first* :sparkles: Final Comment Period! \o/

fitzgen:
FCP complete! The dawn of a new era! :tada:

PR #2, opened 2018-07-12
Proposal to support single inheritance and casting in `wasm-bindgen`

fitzgen:
[Rendered](https://github.com/rustwasm/rfcs/blob/wasm-bindgen-inheritance-casting/text/000-wasm-bindgen-inheritance-casting.md)

### DIFF
--- a/text/000-wasm-bindgen-inheritance-casting.md
+++ b/text/000-wasm-bindgen-inheritance-casting.md
@@ -175,6 +175,15 @@ impl InstanceOf for Whatever {
 }
 ```
 
+`wasm-bindgen` will emit this JavaScript definition of
+`__wbindgen_instanceof_Whatever` that simply wraps the JS `instanceof` operator:
+
+```js
+const __wbindgen_instanceof_Whatever = function (idx) {
+  return getObject(idx) instanceof Whatever;
+};
+```
+
 ## Upcasting Implementation
 
 For every `extends = MyBase` on a type imported with `extern type MyDerived`,
--- a/text/000-wasm-bindgen-inheritance-casting.md
+++ b/text/000-wasm-bindgen-inheritance-casting.md
@@ -73,8 +73,8 @@ interface MyDerived : MyBase {}
 
 ### Example Upcasting
 
-We can upcast into a `MyBase` from a `MyDerived` type using the normal `From`
-and `Into` conversions:
+We can upcast into a `MyBase` from a `MyDerived` type using the normal `From`,
+`AsRef`, `AsMut`, and `Into` conversions:
 
 ```rust
 let derived: MyDerived = get_derived_from_somewhere();
@@ -83,11 +83,13 @@ let base: MyBase = derived.into();
 
 ### Example Dynamically-Checked Casting
 
-We can do dynamically-checked downcasts from a `MyBase` into a `MyDerived`:
+We can do dynamically-checked (checked with JavaScript's `instanceof` operator)
+downcasts from a `MyBase` into a `MyDerived` using the `dyn_{into,ref,mut}`
+methods:
 
 ```
 let base: MyBase = get_base_from_somewhere();
-match base.try_into::<MyDerived>() {
+match base.dyn_into::<MyDerived>() {
     Ok(derived) => {
         // It was an instance of `MyDerived`!
     }
@@ -114,163 +116,131 @@ let derived: MyDerived = base.unchecked_into();
 Unchecked casting serves as an escape hatch for developers, and while it can
 lead to JavaScript exceptions, it cannot create memory unsafety.
 
-## Foundation: The `InstanceOf` Trait
+## The `JsCast` Trait
 
-For dynamically-checked and arbitrary unchecked casting, we introduce the
-`InstanceOf` trait. It provides a boolean predicate that whose implementations
-consult JavaScript's `isntanceof` operator, as well as unchecked conversions
-from JavaScript values.
+For dynamically-checked and unchecked casting between arbitrary JavaScript
+types, we introduce the `JsCast` trait. It requires implementations provide a
+boolean predicate that consults JavaScript's `instanceof` operator, as well as
+unchecked conversions from JavaScript values:
 
 ```rust
-pub trait InstanceOf {
+pub trait JsCast {
     fn instanceof(val: &JsValue) -> bool;
 
     fn unchecked_from_js(val: JsValue) -> Self;
     fn unchecked_from_js_ref(val: &JsValue) -> &Self;
     fn unchecked_from_js_mut(val: &mut JsValue) -> &mut Self;
+
+    // ... provided methods elided ...
 }
 ```
 
-`InstanceOf` is not intended to be used to directly, but instead as a bound on
-type parameters for generic methods on `JsValue` and imported types. Users of
-`wasm-bindgen` will be able to ignore `InstanceOf` for the most part.
+`JsCast`'s required trait methods are not intended to be used to directly, but
+instead are leveraged by its provided methods. Users of `wasm-bindgen` will be
+able to ignore `JsCast`'s required trait methods for the most part, since the
+implementations will be mechanically generated, and they will only be using the
+required trait methods indirectly through the more ergonomic provided methods.
 
-For every `extern type Whatever` imported with `wasm-bindgen`, we emit an
-implementation of `InstanceOf` similar to this:
+For every `extern { type Illmatic; }` imported with `wasm-bindgen`, we emit an
+implementation of `JsCast` similar to this:
 
 ```rust
-impl InstanceOf for Whatever {
+impl JsCast for Illmatic {
     fn instanceof(val: &JsValue) -> bool {
         #[cfg(all(target_arch = "wasm32", not(target_os = "emscripten")))]
         #[wasm_import_module = "__wbindgen_placeholder__"]
         extern {
-            fn __wbindgen_instanceof_Whatever(idx: u32) -> u32;
+            fn __wbindgen_instanceof_Illmatic(idx: u32) -> u32;
         }
 
         #[cfg(not(all(target_arch = "wasm32", not(target_os = "emscripten"))))]
-        unsafe extern fn __wbindgen_instanceof_Whatever(_: u32) -> u32 {
+        unsafe extern fn __wbindgen_instanceof_Illmatic(_: u32) -> u32 {
             panic!("function not implemented on non-wasm32 targets")
         }
 
         __wbindgen_instance_of_MyDerived(val.idx) == 1
     }
 
-    fn unchecked_from_js(val: JsValue) -> Whatever {
-        Whatever {
+    fn unchecked_from_js(val: JsValue) -> Illmatic {
+        Illmatic {
             obj: val,
         }
     }
 
-    fn unchecked_from_js_ref(val: &JsValue) -> &Whatever {
+    fn unchecked_from_js_ref(val: &JsValue) -> &Illmatic {
         unsafe {
-            &*(val as *const JsValue as *const Whatever)
+            &*(val as *const JsValue as *const Illmatic)
         }
     }
 
-    fn unchecked_from_js_mut(val: &mut JsValue) -> &mut Whatever {
+    fn unchecked_from_js_mut(val: &mut JsValue) -> &mut Illmatic {
         unsafe {
-            &mut *(val as *mut JsValue as *mut Whatever)
+            &mut *(val as *mut JsValue as *mut Illmatic)
         }
     }
 }
 ```
 
-`wasm-bindgen` will emit this JavaScript definition of
-`__wbindgen_instanceof_Whatever` that simply wraps the JS `instanceof` operator:
+Additionally, `wasm-bindgen` will emit this JavaScript definition of
+`__wbindgen_instanceof_Illmatic` that simply wraps the JS `instanceof` operator:
 
 ```js
-const __wbindgen_instanceof_Whatever = function (idx) {
-  return getObject(idx) instanceof Whatever;
+const __wbindgen_instanceof_Illmatic = function (idx) {
+  return getObject(idx) instanceof Illmatic;
 };
 ```
 
-## Upcasting Implementation
+### `JsCast`'s Provided Trait Methods
 
-For every `extends = MyBase` on a type imported with `extern type MyDerived`,
-and for every base and derived interface in a WebIDL interface inheritance
-chain, `wasm-bindgen` will emit these trait implementations that wrap unchecked
-conversions methods from `InstanceOf` that we know are valid due to the
-inheritance relationship:
-
-1. A `From` implementation for `self`-consuming conversions:
-
-   ```rust
-   impl From<MyDerived> for MyBase {
-       fn from(my_derived: MyDerived) -> MyBase {
-           let val: JsValue = my_derived.into();
-           <MyDerived as InstanceOf>::unchecked_from_js(val)
-       }
-   }
-   ```
-
-2. An `AsRef` implementation for shared reference conversions:
-
-   ```rust
-   impl AsRef<MyBase> for MyDerived {
-       fn as_ref(&self) -> &MyDerived {
-           let val: &JsValue = self.as_ref();
-           <MyDerived as InstanceOf>::uncheck_from_js_ref(val)
-       }
-   }
-   ```
-
-3. An `AsMut` implementation for exclusive reference conversions:
-
-   ```rust
-   impl AsMut<MyBase> for MyDerived {
-       fn as_mut(&mut self) -> &mut MyDerived {
-           let val: &mut JsValue = self.as_mut();
-           <MyDerived as InstanceOf>::uncheck_from_js_mut(val)
-       }
-   }
-   ```
-
-## The `JsCast` Trait
-
-The `JsCast` trait wraps the `InstanceOf` trait to provide unchecked arbitrary
-casting and dynamically-checked, fallible casting for `JsValue` and imported JS
-classes.
+The `JsCast` trait's provided methods wrap the unergonomic required static trait
+methods and provide ergonomic, chainable versions that operate on `self` and
+*another* `T: JsCast`. For example, the `JsCast::is_instance_of` method asks if
+`&self` is an instance of some *other* `T` that also implements `JsCast`.
 
 ```rust
 pub trait JsCast
 where
     Self: AsRef<JsValue> + AsMut<JsValue> + Into<JsValue>,
 {
-    // Unchecked conversions.
+    // ... required trait methods elided ...
+
+    // Unchecked conversions from `Self` into some other `T: JsCast`.
 
     fn unchecked_into<T>(self) -> T
     where
-        T: InstanceOf,
+        T: JsCast,
     {
         T::unchecked_from_js(self.into())
     }
+
     fn unchecked_ref<T>(&self) -> &T
     where
-        T: InstanceOf,
+        T: JsCast,
     {
         T::unchecked_from_js_ref(self.as_ref())
     }
+
     fn unchecked_mut<T>(&mut self) -> &mut T
     where
-        T: InstanceOf,
+        T: JsCast,
     {
         T::unchecked_from_js_mut(self.as_mut())
     }
 
-    // Dynamic instanceof check.
+    // Predicate method to check whether `self` is an instance of `T` or not.
 
     fn is_instance_of<T>(&self) -> bool
     where
-        T: InstanceOf,
+        T: JsCast,
     {
         T::instanceof(self.as_ref())
     }
 
-    // Dynamically-checked conversions.
+    // Dynamically-checked conversions from `Self` into some other `T: JsCast`.
 
-    fn try_into<T>(self) -> Result<T, Self>
+    fn dyn_into<T>(self) -> Result<T, Self>
     where
-        T: InstanceOf,
+        T: JsCast,
     {
         if self.is_instance_of::<T>() {
             Ok(self.unchecked_into())
@@ -279,9 +249,9 @@ where
         }
     }
 
-    fn try_ref<T>(&self) -> Option<&T>
+    fn dyn_ref<T>(&self) -> Option<&T>
     where
-        T: InstanceOf,
+        T: JsCast,
     {
         if self.is_instance_of::<T>() {
             Some(self.unchecked_ref())
@@ -290,9 +260,9 @@ where
         }
     }
 
-    fn try_mut<T>(&mut self) -> Option<&mut T>
+    fn dyn_mut<T>(&mut self) -> Option<&mut T>
     where
-        T: InstanceOf,
+        T: JsCast,
     {
         if self.is_instance_of::<T>() {
             Some(self.unchecked_mut())
@@ -303,33 +273,26 @@ where
 }
 ```
 
-Using these methods provides better turbo-fishing syntax than using `InstanceOf`
-trait methods directly.
+Using these methods provides better turbo-fishing syntax than using `JsCast`'s
+required trait methods directly.
 
 ```rust
 fn get_it() -> JsValue { ... }
 
-// Wack
+// Tired -_-
 SomeJsThing::unchecked_from_js(get_it()).method();
 
-// Wow! Much chain, very ergo!
+// Wired ^_^
 get_it()
     .unchecked_into::<SomeJsThing>()
     .method();
 ```
 
-### `JsCast` Blanket Implementation
+### `JsCast` Implementation for `JsValue`
 
-We provide a blanket implementation of `JsCast` for anything that is `JsValue`-y
-in that it and its references can be converted into `JsValue`s. This covers all
-`wasm-bindgen`-imported JS classes.
-
-```rust
-impl<T> JsCast for T where T: AsRef<JsValue> + AsMut<JsValue> + Into<JsValue> {}
-```
-
-We also add `AsRef<JsValue>` and `AsMut<JsValue>` implementations for `JsValue`
-itself, so that the blanket implementation applies to `JsValue`:
+We also trivially implement `JsCast` for `JsValue` with no-ops, and add
+`AsRef<JsValue>` and `AsMut<JsValue>` implementations for `JsValue` itself, so
+that the `JsCast` super trait bounds are satisfied:
 
 ```rust
 impl AsRef<JsValue> for JsValue {
@@ -343,7 +306,66 @@ impl AsMut<JsValue> for JsValue {
         self
     }
 }
+
+impl JsCast for JsValue {
+    fn instanceof(_: &JsValue) -> bool {
+        true
+    }
+
+    fn unchecked_from_js(val: JsValue) -> Self {
+        val
+    }
+
+    fn unchecked_from_js_ref(val: &JsValue) -> &Self {
+        val
+    }
+
+    fn unchecked_from_js_mut(val: &mut JsValue) -> &mut Self {
+        val
+    }
+}
 ```
+
+## Upcasting Implementation
+
+For every `extends = MyBase` on a type imported with `extern type MyDerived`,
+and for every base and derived interface in a WebIDL interface inheritance
+chain, `wasm-bindgen` will emit these trait implementations that wrap unchecked
+conversions methods from `JsCast` that we know are valid due to the inheritance
+relationship:
+
+1. A `From` implementation for `self`-consuming conversions:
+
+   ```rust
+   impl From<MyDerived> for MyBase {
+       fn from(my_derived: MyDerived) -> MyBase {
+           let val: JsValue = my_derived.into();
+           <MyDerived as JsCast>::unchecked_from_js(val)
+       }
+   }
+   ```
+
+2. An `AsRef` implementation for shared reference conversions:
+
+   ```rust
+   impl AsRef<MyBase> for MyDerived {
+       fn as_ref(&self) -> &MyDerived {
+           let val: &JsValue = self.as_ref();
+           <MyDerived as JsCast>::uncheck_from_js_ref(val)
+       }
+   }
+   ```
+
+3. An `AsMut` implementation for exclusive reference conversions:
+
+   ```rust
+   impl AsMut<MyBase> for MyDerived {
+       fn as_mut(&mut self) -> &mut MyDerived {
+           let val: &mut JsValue = self.as_mut();
+           <MyDerived as JsCast>::uncheck_from_js_mut(val)
+       }
+   }
+   ```
 
 ## Deep Inheritance Chains Example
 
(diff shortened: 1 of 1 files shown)


icefoxen:
My very un-expert two cents...

I like requiring explicit casting 'cause explicit is good.  Using `From` and `TryFrom` instead feels more Rusty but loses semantic meaning.

Don't re-export the traits in prelude, one can always import them oneself as necessary.  We can always change that later.

`instanceof` (or at least `typeof`) is often necessary in JS code and such, so generating helper functions for it might be nice.

ohanar:
A few first thoughts:

- You don't always just want to upcast or downcast one level at a time for a deeply nested inheritance chain. Using an associated type on Upcast, you can only cast up or down one level at a time, so even in single inheritance I think we should probably have a trait parameter rather than an associated type for Upcast.
- I like the idea of using the standard conversion traits, and I don't really think that having all the extra impls is that much of an issue since everything is just being generated. On the other hand, if you have some a field in a borrowed struct you want to cast, it is a bit annoying and can be confusing to newcomers: `(&foo.bar).into()` vs `foo.bar.upcast_ref()`. Given this, I'm actually leaning towards specialize Upcast/Downcast traits, but it isn't a strong preference.
- Related to the previous point, we could both implement From/TryFrom traits as well Upcast/Downcast. This is already partially suggested in the RFC by including one impl for the From trait (for owned values). I would either go all the way, and implement all the relevant From/TryFrom and Upcast/Downcast, or make a conscious decision to use our own casting traits, and only implement the Upcast/Downcast traits.
- Regarding the `instanceof` helper functions, if it is exposed, we could make it a static method on the type that's being targeted, rather than having a bunch of loose functions. For instance you could call `Foo::has_instance(&bar)` to check if `bar `is an instance of `Foo`. Alternatively or in addition we could make it a method and just use the turbo fish: `bar.instance_of::<Foo>()`.

alexcrichton:
Thanks for the RFC @fitzgen! 

I think may be important here to define in this RFC what role we think these traits and/or mechanisms will have on ecosystems API. On one end of the spectrum we could say that we want the solution here to be as ergonomic as possible, setting the precedent for how the entire ecosystem uses JS objects and casts. On the other end we could say that we only want to support the ability to connect types between their parent and child classes, but don't care about ergonomics at all.

I personally feel that we should probably lean towards the latter end of the spectrum, mainly *empowering* users here rather than defining ecosystem-wide conventions about how they're expected to be used. The `web-sys` crate, like `libc`, is intended to be the foundation of many crates. The `libc` crate, for example, doesn't attempt to set any trends about how to use APIs, exporting `unsafe extern fn getpid() -> c_int` which neither needs to be `unsafe` nor really needs to return a `c_int` as an extra layer of indirection (for example).

Along those lines I think I'd like to propose another alternative to this RFC. I feel that the `Upcast` and `Downcast` traits introduced here sort of sit in the middle of the spectrum mentioned above. One one hand I don't think they're suitable for ecosystem-wide ergonomics because I don't think you can define a function that takes any subclass of `Node`, and additionally for downcasting it doesn't currently (although it could certainly have them) have unchecked downcasts.

The alternative I'm thinking of [looks like this](https://play.rust-lang.org/?gist=fe67d276318809be30500f1dfc2d6d48&version=stable&mode=debug&edition=2015):

```rust
pub unsafe trait AllJsValues {
    fn from_js_value(js: JsValue) -> Self where Self: Sized;
    fn into_js_value(self) -> JsValue where Self: Sized;
    
    fn cast<U>(self) -> U
        where U: AllJsValues + Sized, Self: Sized 
    {
        U::from_js_value(self.into_js_value())
    }
        
    fn cast_ref<U>(&self) -> &U
        where U: AllJsValues
    {
        unsafe { &*(self as *const Self as *const U) }
    }

    fn cast_mut<U>(&mut self) -> &mut U
        where U: AllJsValues
    {
        unsafe { &mut *(self as *mut Self as *mut U) }
    }
}
```

An implementation of this trait could be generated by `#[wasm_bindgen]` for all imported types. I see this version of the trait on the extreme end of "let's provide low level functionality and that's it". The idea here is that it allows you to cast between any types anywhere (as that's sort of in theory what you can do in JS). It still leaves room, however, for a perhaps more formal abstraction in the ecosystem.

We could also be even more "extreme" and purely provide accessors for the `JsValue`, `&JsValue`, and `&mut JsValue` internally, then providing constructors for all three as well for all types, meaning casts have to go into some sort of a `JsValue` and then get casted into a different type. (this could all be `From` traits I believe?)

In any case I'm personally a fan of sticking to the lower end of the spectrum of "not providing too much abstraction", but I'm curious what others think as well!

fitzgen:
@ohanar 

> You don't always just want to upcast or downcast one level at a time for a deeply nested inheritance chain. Using an associated type on Upcast, you can only cast up or down one level at a time, so even in single inheritance I think we should probably have a trait parameter rather than an associated type for Upcast.

This is a really good insight!

My one fear is that we lose the type inference that allows chaining like

```rust
derived.upcast().base_method();
```

and instead are forced to write temporaries and help out type inference:

```rust
let base: MyBase = derived.upcast();
base.base_method();
```

I wonder if we can separate facts about inheritance relationships from the methods that do the casting so that we can maintain ergonomics. I think turbo-fishing is the best we could do syntax/ergonomics-wise with non-linear upcasting:

```rust
derived.upcast::<MyBase>().base_method();
```

This would require trait definitions like

```rust
pub unsafe trait Extends<T>: Sized + Into<T> {}

pub trait Upcast {
    fn upcast<T>(self) -> T
    where
        Self: Extends<T>;

    fn upcast_ref<T>(&self) -> &T
    where
        Self: Extends<T>;

    fn upcast_mut<T>(&mut self) -> &mut T
    where
        Self: Extends<T>;
}

impl<U> Upcast for U {
    #[inline]
    fn upcast<T>(self) -> T
    where
        Self: Extends<T>,
    {
        self.into()
    }

    #[inline]
    fn upcast_ref<T>(&self) -> &T
    where
        Self: Extends<T>,
    {
        unsafe { std::mem::transmute(self) }
    }

    #[inline]
    fn upcast_mut<T>(&mut self) -> &mut T
    where
        Self: Extends<T>,
    {
        unsafe { std::mem::transmute(self) }
    }
}
```

So basically it comes down to whether non-linear upcasting is worth having to turbofish in method chaining like `derived.upcast::<MyBase>().base_method()`.

Thoughts?

[(Playground with this version of the upcasting trait)](https://play.rust-lang.org/?gist=cb1fdb8531970a2e167f3be3e6a7dcc6&version=stable&mode=debug&edition=2015)

fitzgen:
@alexcrichton

> I think may be important here to define in this RFC what role we think these traits and/or mechanisms will have on ecosystems API. On one end of the spectrum we could say that we want the solution here to be as ergonomic as possible, setting the precedent for how the entire ecosystem uses JS objects and casts. On the other end we could say that we only want to support the ability to connect types between their parent and child classes, but don't care about ergonomics at all.
>
> I personally feel that we should probably lean towards the latter end of the spectrum, mainly empowering users here rather than defining ecosystem-wide conventions about how they're expected to be used. The `web-sys` crate, like `libc`, is intended to be the foundation of many crates. The `libc` crate, for example, doesn't attempt to set any trends about how to use APIs, exporting `unsafe extern fn getpid() -> c_int` which neither needs to be unsafe nor really needs to return a `c_int` as an extra layer of indirection (for example).

Agreed that our number one goal is empowering users to leverage APIs that make use of inheritance, that ergonomics is a nice-to-have, and that setting conventions for all crates built on top of `wasm-bindgen` is a non-goal.

However, I would like to disentangle type safety from ergonomics.

`libc` is not type safe, and therefore raw, thin bindings to it are also not type safe.

On the other hand, Web APIs/WebIDL *are* type safe, and we shouldn't *lose* that property when writing raw, thin bindings to them.

I'm happy to include `unsafe fn unchecked_cast<T: Into<JsValue>>(T) -> Self` methods on imported types, but I don't think that alone is sufficient.

So looking at the problem once again from the perspective of

* type-safe casting as a hard requirement, and
* ergonomics of that casting as nice-to-have only when it comes "free"

then I don't think `AllJsValues` is a solution. But I think both the RFC's current `Upcast` and the `Upcast` from my last comment both fit these goals.

alexcrichton:
Excellent points @fitzgen! I think it's worth digging into the type safety point here. I personally think we should be wary, though, of ascribing type safety to the synthetic types we're attaching to JS objects. 

We've got WebIDL for the web, but how sure are we that it's 100% exactly what all browsers accept? Are we sure that there aren't at least a few corner cases of some "duck typed" APIs or otherwise instances of accepting more than one type for compatibility's sake?

Additionally, while WebIDL may be typesafe I'd imagine that the NPM ecosystem at large likely isn't in the sense that some types could be ascribed but at least in Typescript what I've seen is a lot of interfaces rather than classes (like a trait object instead of a concrete type) and there's not a great way to express that I think?

I think what I'm trying to say is that it feels to me like an already-lost uphill battle to ascribe types to the JS ecosystem with 100% fidelity in the sense that we can use `unsafe` for "you've basically created a segfault with this API". I would instead expect this to be 80-90% fidelity (or something like that) and there are cases where you'd need to cheat a bit with types. If we ascribe Rust's `unsafe` to this sort of "required cheating" it feels to me like a misuse of `unsafe` in the sense that `unsafe` is used to demarcate issues that can be caught with 100% fidelity.

fitzgen:
> I think what I'm trying to say is that it feels to me like an already-lost uphill battle to ascribe types to the JS ecosystem with 100% fidelity in the sense that we can use unsafe for "you've basically created a segfault with this API". I would instead expect this to be 80-90% fidelity (or something like that) and there are cases where you'd need to cheat a bit with types.

This doesn't seem to match with `wasm-bindgen`'s existing philosophy:

* We already statically type imported functions, instead of only returning `JsValue`s and dynamically type checking them all over.

* We already make distinctions between functions that can throw and functions that can't, instead of catching every single time we call into JS and requiring that every imported function return `Result<_, JsValue>`.

If we aren't going to trust our input (webidl / proc-macro attributes) and the types described therein, then I don't think we should be returning typed values from imported functions nor trusting whether an imported function may throw or not. In this world where we don't trust JS types, we should apply the "be conservative in what you send, be liberal in what you accept" principal to **all** interactions with JS:

*  For imported functions and methods, we could statically type parameters, but we could only receive `Result<JsValue, JsValue>`s as return types (either a normal return value or a thrown value).

* For exported functions, we would only accept `JsValue` parameters, but we could return more specific, constrained types.

FWIW, I don't think this would be a bad world to live in. But it is very different from what we are doing now.

> If we ascribe Rust's unsafe to this sort of "required cheating" it feels to me like a misuse of unsafe in the sense that unsafe is used to demarcate issues that can be caught with 100% fidelity.

The whole concept of `unsafe` is a bit funky when living inside a safe runtime. I think it makes more sense to talk about leveraging types. I would be happy to compromise on `unchecked_cast` not being `unsafe` -- I don't think it really matters that much in the wasm context.

alexcrichton:
Hm so I definitely don't mean to say that we should distrust all the webidl sources. Nor do I think it's a lost cause to *attempt* to add types here and there to the web. Mainly though what I'm trying to say is that we shouldn't be striving for perfection. That may be a bit of a nebulous concept in my head though, but to me `unsafe fn unchecked_cast(...)` is an example of "perfection" where we're declaring we got everything "right enough" where you should basically almost never have to use that sort of function.

I do think there's a lot of value in adding types *somewhere* in the ecosystem and I think putting it at the `web-sys` layer is pretty appropriate. I believe, though, that the apis should be pretty flexible in what they accept. The types there are sort of for what 95% of the API calls accept but the remaining 5%+ would be done with casts and such as necessary.

This feels similar to the standard library's `File` type for example in that the low-level APIs all just take numbers and pointers, but we add a type wrapped around that, `File`, which is stronger than it needs to be. It's what you need 90% of the time though so we have the `IntoRawFd` and `FromRawFd` traits as an escape hatch if necessary (the `unsafe` here is for weird other reasons).

I guess what I'm getting at is that a perhaps even more conservative implementation of this RFC (to the end of "expose all runtime capabilities") is something like:

```rust
#[wasm_bindgen]
extern {
    type Foo;
}

// becomes ...

struct Foo(JsValue);

impl AsRef<JsValue> for Foo { ... }
impl AsMut<JsValue> for Foo { ... }
impl From<JsValue> for Foo { ... }
impl AsRef<Foo> for JsValue { ... }
impl AsMut<Foo> for JsValue { ... }
impl From<Foo> for JsValue { ... }
```

That way you should be able to convert in every direction as well as making a super-general API take something like `T: AsRef<JsValue>`.

Pauan:
> I guess what I'm getting at is that a perhaps even more conservative implementation of this RFC (to the end of "expose all runtime capabilities") is something like:

That's basically what stdweb does. Here's how you define a new JS type with stdweb:

```rust
#[derive(Clone, PartialEq, Eq, ReferenceType)]
#[reference(instance_of = "RegExp")]
pub struct RegExp(Reference);
```

The above code gets expanded (by the derive macros) into something like this:

```rust
pub struct RegExp(Reference);

impl AsRef<Reference> for RegExp { ... }
impl From<RegExp> for Reference { ... }
impl TryFrom<Reference> for RegExp { ... }
impl TryFrom<Value> for RegExp { ... }
impl JsSerialize for RegExp { ... }
impl InstanceOf for RegExp { ... }
```

* The [`Value`](https://docs.rs/stdweb/0.4.7/stdweb/enum.Value.html) type is for all JS values.

* The [`Reference`](https://docs.rs/stdweb/0.4.7/stdweb/struct.Reference.html) type is for JS objects only.

* The [`JsSerialize`](https://docs.rs/stdweb/0.4.7/stdweb/trait.JsSerialize.html) trait is for things which can be sent to JS.

* The [`InstanceOf`](https://docs.rs/stdweb/0.4.7/stdweb/trait.InstanceOf.html) trait adds automatic run-time `instanceof` checking when converting a JS value into Rust (to ensure correctness).

This works out well in practice with stdweb, so I agree with your idea of making everything generic.

----

Speaking of inheritance, stdweb solves that with two systems:

1. Method inheritance is handled with traits. As an example:

   1. The [`IHtmlElement`](https://docs.rs/stdweb/0.4.7/stdweb/traits/trait.IHtmlElement.html) trait contains all of the methods for HTMLElement.

   2. `IHtmlElement` inherits from the [`IElement`](https://docs.rs/stdweb/0.4.7/stdweb/web/trait.IElement.html) trait which contains all the methods for Element.

   3. `IElement` inherits from the [`INode`](https://docs.rs/stdweb/0.4.7/stdweb/web/trait.INode.html), [`IParentNode`](https://docs.rs/stdweb/0.4.7/stdweb/web/trait.IParentNode.html), and [`IChildNode`](https://docs.rs/stdweb/0.4.7/stdweb/web/trait.IChildNode.html) traits (this "multiple inheritance" is required by WebIDL because it has mixins/interfaces).

2. In addition to the traits, there are also structs which contain the actual JS objects (for example there's the `IHtmlElement` trait and the [`HtmlElement`](https://docs.rs/stdweb/0.4.7/stdweb/web/struct.HtmlElement.html) struct, then the `IElement` trait and the [`Element`](https://docs.rs/stdweb/0.4.7/stdweb/web/struct.Element.html) struct, etc.)

   Object inheritance is handled with `From` and `TryFrom`. When you define a type, you can mark it as a sub-class of other types, like this:

   ```rust
   #[derive(Clone, Debug, PartialEq, Eq, ReferenceType)]
   #[reference(instance_of = "HTMLElement")]
   #[reference(subclass_of(EventTarget, Node, Element))]
   pub struct HtmlElement(Reference);

   // Implement the various traits for HtmlElement
   impl IEventTarget for HtmlElement {}
   impl INode for HtmlElement {}
   impl IElement for HtmlElement {}
   impl IHtmlElement for HtmlElement {}
   ```

   The important part is the `subclass_of` macro, which automatically generates these impls:

   ```rust
   impl From<HtmlElement> for EventTarget { ... }
   impl TryFrom<EventTarget> for HtmlElement { ... }

   impl From<HtmlElement> for Node { ... }
   impl TryFrom<Node> for HtmlElement { ... }

   impl From<HtmlElement> for Element { ... }
   impl TryFrom<Element> for HtmlElement { ... }
   ```

   As you can see, upcasting is handled with `From` (which always succeeds), whereas downcasting is handled with `TryFrom` (which can fail).

If you look at the methods for the traits, you'll see that they usually accept generic arguments (`IHtmlElement`, `IElement`, etc.)

In other words, they accept anything that implements those traits, so you rarely need to use `From`/`TryFrom` to do casting, which makes things a lot more ergonomic.

fitzgen:
@alexcrichton

> Hm so I definitely don't mean to say that we should distrust all the webidl
> sources. Nor do I think it's a lost cause to attempt to add types here and
> there to the web. Mainly though what I'm trying to say is that we shouldn't be
> striving for perfection.

I think we are in agreement here.

> That may be a bit of a nebulous concept in my head though, but to me unsafe fn
> unchecked_cast(...) is an example of "perfection" where we're declaring we got
> everything "right enough" where you should basically almost never have to use
> that sort of function.

I think the `web-sys` crate's WebIDL-derived definitions will be "right enough"
95% of the time, and one will be able to mostly use static upcasting.

I am not against unchecked dynamic casts, and I am not against marking them safe
either.

> ```rust
> impl From<JsValue> for Foo { ... }
> impl AsRef<Foo> for JsValue { ... }
> impl AsMut<Foo> for JsValue { ... }
> ```

However, I think that these particular implementations are a little too much of
a footgun and encourage bad habits.

If a static upcast *can* be used, then it should be preferred over an unchecked
dynamic cast.

If a static upcast *cannot* be used, then one should have a small road bump to
think about whether they are using a Web API correctly and whether they should
consider an `instanceof` check or not. But they should definitely still have the
option to say "trust me, I got this" and do an unchecked dynamic cast, and that
operation can be safe.

Ultimately, I just want these implementations to be of different traits that are
structurally identical but give users that pause and cause them to think about
what they are doing. They should also have documentation about potential
pitfalls.

```rust
// Unchecked (but still safe) conversions.
impl UncheckedFrom<JsValue> for Foo { ... }
impl UncheckedAsRef<Foo> for JsValue { ... }
impl UncheckedAsMut<Foo> for JsValue { ... }

// Checked conversions using `instanceof`
impl TryFrom<JsValue> for Foo { ... }
impl<'a> TryFrom<&'a JsValue> for &'a Foo { ... }
impl<'a> TryFrom<&'a mut JsValue> for &'a mut Foo { ... }
```

--------------------------------------------------------------------------------

@Pauan

> Here's how you define a new JS type with stdweb:
>
> ```rust
> #[derive(Clone, PartialEq, Eq, ReferenceType)]
> #[reference(instance_of = "RegExp")]
> pub struct RegExp(Reference);
> ```
>
> The above code gets expanded (by the derive macros) into something like this:
>
> ```rust
> pub struct RegExp(Reference);
>
> impl AsRef<Reference> for RegExp { ... }
> impl From<RegExp> for Reference { ... }
> impl TryFrom<Reference> for RegExp { ... }
> impl TryFrom<Value> for RegExp { ... }
> impl JsSerialize for RegExp { ... }
> impl InstanceOf for RegExp { ... }
> ```

I'm assuming these `TryFrom` implementations are doing `instanceof` checks? This
seems a bit different from the *unchecked* casts that @alexcrichton is
proposing.

> 1. Method inheritance is handled with traits. As an example:
>
>    [...]
>
> 2. In addition to the traits, there are also structs which contain the actual JS
>    objects (for example there's the IHtmlElement trait and the HtmlElement
>    struct, then the IElement trait and the Element struct, etc.)

I alluded to this in the original RFC text. I was hoping that we could do this
sort of thing as a follow up, and implement just the casting incrementally.

I do think this general approach is the best way to get nicest ergonomics we can
out of mechanically-generated bindings. (Hand-written libraries customized for
ergo in some specific context can always do better -- think something like
`glium` -- but I believe this is outside `web-sys`'s responsibilities.)

alexcrichton:
@fitzgen that sounds good to me yeah! It also looks suspiciously similar to the current RFC so I'm trying to reread my original response and the RFC to figure out the main differences... I think it basically just comes down to the details of how we expect these to be used and exactly how we're setting up the traits? 

For example these traits probably aren't suitable for "use everywhere as the base abstraction". An example of that you can't necessarily take `T: Downcast<Base = MyRelevantType>` because it doesn't apply to sub-sub classes. Additionally the `T: Upcast<Self>` on `Downcast` seems a little suspect to me, and it also doesn't provide facilities for downcasting through multiple class chains w/o doing multiple calls I think?

Ok so to may be put my thoughts a bit more clearly, I think I'm worried that the RFC as-written isn't achieving the goal of "expose possibility, even if it's not the most ergonomic". Now I think it can with a few tweaks though! Perhaps something like:

```rust
// in wasm_bindgen crate

pub unsafe trait InstanceOf {
    fn is_instance_of(val: &JsValue) -> bool;
    fn unchecked_from_js(val: JsValue) -> Self;
    fn unchecked_from_js_ref(val: &JsValue) -> &Self;
    fn unchecked_from_js_mut(val: &mut JsValue) -> &mut Self;
}

impl JsValue {
    pub fn instanceof<T>(&self) -> bool {
        T::is_instance_of(self)
    }

    pub fn try_into<T>(self) -> Result<T, Self>
        where T: InstanceOf 
    {
        if self.instanceof::<T>() { Ok(self.unchecked_into()) } else { Err(self) }
    }

    pub fn try_ref<T>(&self) -> Option<&T> 
        where T: InstanceOf
    {
        if self.instanceof::<T>() { Some(self.unchecked_ref()) } else { None }
    }

    pub fn try_mut<T>(&mut self) -> Option<&mut T> 
        where T: InstanceOf
    {
        if self.instanceof::<T>() { Some(self.unchecked_mut()) } else { None }
    }

    pub fn unchecked_into<T>(self) -> T where T: InstanceOf {
        T::unchecked_from_js(self)
    }
    
    pub fn unchecked_ref<T>(&self) -> &T where T: InstanceOf {
        T::unchecked_from_js_ref(self)
    }

    pub fn unchecked_mut<T>(&mut self) -> &mut T where T: InstanceOf {
        T::unchecked_from_js_mut(self)
    }

    // pub fn from<T: Into<Self>>(t: T) -> JsValue { t.into() } // provided by the Rust prelude
    pub fn from_ref<T: AsRef<Self>>(t: &T) -> &JsValue { t.as_ref() }
    pub fn from_mut<T: AsMut<Self>>(t: &mut T) -> &mut JsValue { t.as_mut() }
}

// and then in your crate

#[wasm_bindgen]
type Foo;
#[wasm_bindgen(extends = Foo)]
type Bar;

// generates

impl InstanceOf for Foo {
    fn is_instance_of(val: &JsValue) -> bool { /* shell out to JS shim */ }
    fn unchecked_from_js(val: JsValue) -> Self { /* do the constructor */ }
    fn unchecked_from_js_ref(val: &JsValue) -> &Self { /* do the weird transmute */ }
    fn unchecked_from_js_mut(val: &mut JsValue) -> &mut Self { /* do the weird transmute */ }
}
impl From<Foo> for JsValue { ... }
impl AsRef<JsValue> for Foo { ... }
impl AsMut<JsValue> for Foo { ... }

impl InstanceOf for Bar {
    fn is_instance_of(val: &JsValue) -> bool { /* shell out to JS shim */ }
    fn unchecked_from_js(val: JsValue) -> Self { /* do the constructor */ }
    fn unchecked_from_js_ref(val: &JsValue) -> &Self { /* do the weird transmute */ }
    fn unchecked_from_js_mut(val: &mut JsValue) -> &mut Self { /* do the weird transmute */ }
}

impl From<Bar> for Foo { /* use unchecked casts in InstanceOf */ }
impl AsRef<Foo> for Bar { /* use unchecked casts in InstanceOf */ }
impl AsMut<Foo> for Bar { /* use unchecked casts in InstanceOf */ }
impl From<Bar> for JsValue { ... }
impl AsRef<JsValue> for Bar { ... }
impl AsMut<JsValue> for Bar { ... }
```

...

Hm so actually that's very similar to my [earlier comment](https://github.com/rustwasm/rfcs/pull/2#issuecomment-404629766), but in writing it up I've realized a few things:

* It should be possible to have a zero-cost unchecked cast between any two types
* It should be possible to have a checked cast between any two types using at most one `instanceof`
* It should be possible to upcast up an inheritance chain in a zero-cost fashion with no "unchecked" operations

I'm curious if you agree w/ these constraints? I think it fits what we've been saying here pretty well so far, but want to make sure I didn't miss something!

Now on a totally unrelated note from *functionality* when we get to API-design business I'm personally wary to lean too heavily on traits. I find they can often serve as a barrier to understanding because you're always feeling like you should be using a trait somewhere or are otherwise too aggressively implementing traits on things that aren't used all that often (or bemoaning a missing blanket impl or lack of specialization or something like that). The ergonomics also aren't too great as you have to import them, but `wasm_bindgen`'s prelude can solve this. In any case I liked the idea of using standard Rust prelude traits (if we can) and otherwise discouraging use of the trait here (`InstanceOf`) through APIs on `JsValue` which are slightly more easy to discover/document. This isn't related to the functionality of the API though! Just the exact construction of what's a trait and what isn't.

fitzgen:
> * It should be possible to have a zero-cost unchecked cast between any two types
> * It should be possible to have a checked cast between any two types using at most one instanceof
> * It should be possible to upcast up an inheritance chain in a zero-cost fashion with no "unchecked" operations

Yes! This is the crux of it.

fitzgen:
> * It should be possible to have a zero-cost unchecked cast between any two types
> * It should be possible to have a checked cast between any two types using at most one instanceof
> * It should be possible to upcast up an inheritance chain in a zero-cost fashion with no "unchecked" operations

Yes! This is the crux of it.

fitzgen:
The one question still unanswered is whether we will generate `As{Ref,Mut}` and `From` impls for the whole inheritance chain, or just one level deep (and requiring users to chain `.as_ref()`s).

```rust
#[wasm_bindgen]
extern {
    type Base;

    #[wasm_bindgen(extends = Base)]
    type Derived;

    #[wasm_bindgen(extends = Derived)]
    type DoubleDerived;
}
```

I think for the proc-macro we can only do one level due to technical reasons (for example if the definitions are split across `extern` blocks).

For WebIDL we could possibly do the whole thing.

alexcrichton:
> I think for the proc-macro we can only do one level due to technical reasons 

Agreed!

> For WebIDL we could possibly do the whole thing.

Also agreed!

To bridge the gap we could perhaps allow multiple `extends` attributes for handwritten bindings?

fitzgen:
> To bridge the gap we could perhaps allow multiple extends attributes for handwritten bindings?

SGTM!

Will update the RFC text tomorrow.

fitzgen:
RFC text updated!

@alexcrichton @icefoxen @ohanar what do y'all think?

fitzgen:
@Pauan interested in your feedback on the updated RFC as well! Sorry I forgot to tag you in that last comment :-p

alexcrichton:
Looks great to me, thanks for updating @fitzgen!

Do we need a split between the two traits still? I think it'd be possible to have only one trait for these operations, right?

fitzgen:
> Do we need a split between the two traits still? I think it'd be possible to have only one trait for these operations, right?

This would require implementing `InstanceOf` for `JsValue`, which we *could* do but also makes the `InstanceOf` trait not map directly onto the `instanceof` operator since there is no constructor for js values to use with `instanceof` and it is trivially true that all JS values are instances of JS value.

Thoughts?

alexcrichton:
Oh I think that's fine, it just means that the "instanceof" check there always returns `true` in that everything's a instance of `JsValue`, which seems fine by me!

ohanar:
Overall, I'm pretty happy with the proposal, there are just a couple small comments I have:
- It is a little annoying right now that upcasting with a turbo-fish requires using `unchecked_*`. This might not be too big of an issue after we decide on a more ergonomic way to expose base methods.
- I would prefer avoiding using the name `try_into` in the `JsCast` trait. I would like to leave open the future possibility of implementing the standard fallible conversion traits, and it would be unfortunate to have a name collision.

Pauan:
> @fitzgen I'm assuming these TryFrom implementations are doing instanceof checks?

Yes, the `TryFrom` implementations use `instanceof`. To be more specific, the `TryFrom` implementation calls the [`downcast`](https://docs.rs/stdweb/0.4.8/stdweb/struct.Reference.html#method.downcast) method, which in turn calls the [`instance_of`](https://docs.rs/stdweb/0.4.8/stdweb/trait.InstanceOf.html#tymethod.instance_of) method (which is automatically created by `#[reference(instance_of = "Foo")]`). All of this is automated by the macros, so users don't need to deal with it.

On the other hand, the `From` implementations **don't** use `instanceof` (they are zero-cost and unchecked).

And it's always possible to use [`from_reference_unchecked`](https://docs.rs/stdweb/0.4.8/stdweb/trait.ReferenceType.html#tymethod.from_reference_unchecked) to do an *unsafe* (but zero-cost) cast between different types.

> I alluded to this in the original RFC text. I was hoping that we could do this sort of thing as a follow up, and implement just the casting incrementally.

Sure, I did see that. But since stdweb has already tackled this problem, I wanted to explain the solution they came up with, since it acts as evidence that the general idea presented in this RFC is both sound and practical.

----

> interested in your feedback on the updated RFC as well! Sorry I forgot to tag you in that last comment :-p

Sure! I'll reply individually to parts of the RFC:

> Foundation: The InstanceOf Trait

We should consider marking the `unchecked` methods as `unsafe` (even though they probably can't cause *Rust* unsafety, they still shouldn't be *easy* to use!)

> The JsCast Trait

Why the separation between `InstanceOf` and `JsCast`?

> Only implement its methods on JsValue and require that conversions like ImportedJsClassUno -> ImportedJsClassDos go to JsValue in between: ImportedJsClassUno -> JsValue -> ImpiortedJsClassDos.

I'm fine with this. In almost every case there will be `From` and `TryFrom` (or whatever) implementations (which are auto-generated by `#[wasm_bindgen(extends = Foo)]`), so that sort of manual conversion (with `JsValue` in-between) should be *extremely* rare.

> Should the JsCast trait be re-exported in wasm_bindgen::prelude?

I view manual casting as a very advanced thing which should rarely be needed: users should use `From` / `TryFrom` instead. So I **don't** think `JsCast` should be exported in the prelude.

fabricedesre:
One use case I have is to implement Custom Elements in Rust. In this situation, the Rust struct needs to inherit from (for instance) HTMLElement and implement the custom elements lifecycle callbacks (see details at https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements#Using_the_lifecycle_callbacks).

It's unclear to me how this RFC would allow that since it seems focused on exposing the JS inheritance chain to Rust when importing JS but not when exporting.

alexcrichton:
@fabricedesre I think that's covered by https://github.com/rustwasm/wasm-bindgen/issues/210 which isn't addressed by this RFC, although the components in this RFC should be hopefully adding support for it in the future!

fitzgen:
New commit merged `InstanceOf` into the `JsCast` trait, so there is only one trait here.

We implement `JsCast` for `JsValue` with trivial no-ops and identity functions.

Renamed `JsCast::try_{into,ref,mut}` to `JsCast::dyn_{into,ref,mut}` so that we are future compatible with `TryFrom` and `TryInto`.

fitzgen:
I'd like to now propose that we enter the Final Comment Period for this RFC.

*Aside: we might want to eventually make a bindings/lib/something team that handles RFCs that are related to wasm-bindgen's output and design, but since such a team does not exist at the moment, I will ask @rustwasm/core to sign off.*

Disposition: **merge**

* [X] @fitzgen
* [x] @ashleygwilliams 
* [x] @alexcrichton

fitzgen:
All @rustwasm/core team members have signed off. Entering the 7 day :sparkles: Final Comment Period :sparkles:!

alexcrichton:
A sample implementation of the current state of this RFC is at https://github.com/rustwasm/wasm-bindgen/pull/640

fitzgen:
FCP complete! Thanks everyone for the discussion :) I think we ended up in a better place than where we started!

PR #8, opened 2019-02-15
RFC: Enable depending on NPM packages

alexcrichton:
This proposal is an evolution of https://github.com/rustwasm/rfcs/pull/4 which accounts for much of the discussion and makes a concrete proposal. 

[Rendered](https://github.com/alexcrichton/rfcs-1/blob/npm-deps/text/008-npm-dependencies.md)

### DIFF
--- a/text/008-npm-dependencies.md
+++ b/text/008-npm-dependencies.md
@@ -43,7 +43,7 @@ requirement, need to be declare. This RFC proposes doing this in a
 ```json
  {
   "dependencies": {
-    "foo": "1.0.1"
+    "foo": "^1.0.1"
   }
 }
 ```
--- a/text/008-npm-dependencies.md
+++ b/text/008-npm-dependencies.md
@@ -1,0 +1,218 @@
+- Start Date: 2018-02-14
+- RFC PR: (leave this empty)
+- Tracking Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Enable Rust crates to transparently depend on packages in the npm ecosystem.
+These dependencies will, like normal Rust dependencies through Cargo, work
+seamlessly when consumed by other crates.
+
+# Motivation
+[motivation]: #motivation
+
+The primary goal of `wasm-bindgen` and `wasm-pack` is to enable seamless
+integration of Rust with JS. A massive portion of the JS ecosystem, npm, however
+currently has little support in `wasm-bindgen` and `wasm-pack`, making it
+difficult to access this rich resource that JS offers!
+
+The goal of this RFC is to enable these dependencies to exist. Rust crates
+should be able to require functionality from NPM, just like how NPM can require
+Rust crates compiled to wasm. Any workflow which currently uses NPM packages
+(such as packaging WebAssembly with a bundler) should continue to work but also
+allow pulling in "custom" NPM packages as well as requested by Rust
+dependencies.
+
+# Stakeholders
+[stakeholders]: #stakeholders
+
+This RFC primarily affects uses of `wasm-pack` and `wasm-bindgen` who are also
+currently using bundlers like Webpack. This also affects, however, developers of
+core foundational crates in the Rust ecosystem who want to be concious of the
+ability to pull in NPM dependencies and such.
+
+# Detailed Explanation
+[detailed-explanation]: #detailed-explanation
+
+Adding an NPM dependency to a Rust project will look very similar to adding an
+NPM dependency to a normal JS project. First the dependency, and its version
+requirement, need to be declare. This RFC proposes doing this in a
+`package.json` file adjacent to the crate's `Cargo.toml` file:
+
+```json
+ {
+  "dependencies": {
+    "foo": "1.0.1"
+  }
+}
+```
+
+The `package.json` file will initially be a subset of NPM's `package.json`,
+only supporting one `dependencies` top-level key which internally has key/value
+pairs with strings. Beyond this validation though no validation will be
+performed of either key or value pairs within `dependencies`. In the future
+it's intended that more keys of `package.json` in NPM will be supported, but
+this RFC is intended to be an MVP for now to enable dependencies on NPM at all.
+
+After this `package.json` file is created, the package next needs to be
+imported in the Rust crate. Like with other Rust dependencies on JS, this will
+be done with the `#[wasm_bindgen]` attribute:
+
+```rust
+#[wasm_bindgen(module = "foo")]
+extern "C" {
+    fn function_in_foo_package();
+}
+```
+
+> **Note**: in JS the above import would be similar to:
+>
+> ```js
+> import { function_in_foo_package } from "foo";
+> ```
+
+The exiting `module` key in the `#[wasm_bindgen]` attribute can be used to
+indicate which ES module the import is coming from. This affects the `module`
+key in the final output wasm binary, and corresponds to the name of the package
+in `package.json`. This is intended to match how bundler conventions already
+interpret NPM packages as ES modules.
+
+After these two tools are in place, all that's needed is a `wasm-pack build` and
+you should be good to go! The final `package.json` will have the `foo`
+dependency listed in our `package.json` above and be ready for consumption via a
+bundler.
+
+### Technical Implementation
+
+Under the hood there's a few moving parts which enables all of this to happen.
+Let's first take a look at the pieces in `wasm-bindgen`.
+
+The primary goal of this RFC is to enable *tranparent* and *transitive*
+dependencies on NPM. The `#[wasm_bindgen]` macro is the only aspect of a crate's
+build which has access to all transitive dependencies, so this is what we'll be
+using to slurp up `package.json`. When `#[wasm_bindgen]` with a `module` key is
+specified it will look for `package.json` inside the cwd of the procedural macro
+(note that the cwd is set by Cargo to be the directory with the crate's
+`Cargo.toml` that is being compiled, or the crate in which `#[wasm_bindgen]` is
+written). This `package.json`, if found, will have an absolute path to it
+encoded into the custom section that `wasm-bindgen` already emits.
+
+Later, when the `wasm-bindgen` CLI tool executes, it will parse an interpret all
+items in the wasm-bindgen custom section. All `package.json` files listed will
+be loaded, parsed, and validated (aka only `dependencies` allowed for now). If
+any `package.json` is loaded then a `package.json` file will be emitted next to
+the output JS file inside of `--out-dir`.
+
+After `wasm-bindgen` executes, then `wasm-pack` will read the `package.json`
+output, if any, and augment it with metadata and other items which are already
+emitted.
+
+If more than one crate in a dependency graph depends on an NPM package then in
+this MVP proposal an error will be generated. In the future we can implement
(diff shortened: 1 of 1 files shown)


text/008-npm-dependencies.md
```diff
@@ -0,0 +1,218 @@
...
+
+### Technical Implementation
+
+Under the hood there's a few moving parts which enables all of this to happen.
+Let's first take a look at the pieces in `wasm-bindgen`.
+
+The primary goal of this RFC is to enable *tranparent* and *transitive*
+dependencies on NPM. The `#[wasm_bindgen]` macro is the only aspect of a crate's
+build which has access to all transitive dependencies, so this is what we'll be
+using to slurp up `package.json`. When `#[wasm_bindgen]` with a `module` key is
```
fitzgen, line 94: Nice insight.

text/008-npm-dependencies.md
```diff
@@ -0,0 +1,218 @@
...
+specified it will look for `package.json` inside the cwd of the procedural macro
+(note that the cwd is set by Cargo to be the directory with the crate's
+`Cargo.toml` that is being compiled, or the crate in which `#[wasm_bindgen]` is
+written). This `package.json`, if found, will have an absolute path to it
+encoded into the custom section that `wasm-bindgen` already emits.
```
fitzgen, line 99: This dependency on package.json will be opaque to `cargo` -- does it make sense to always include the `package.json` path (even when it doesn't exist) in the custom section and have the CLI tool handle the missing case? This way incremental builds where a `package.json` is included post facto should always work.

I think modifications to package.json would work incrementally either way though, right?

text/008-npm-dependencies.md
```diff
@@ -0,0 +1,218 @@
...
+to generate code that's compatible with these locations of hosting NPM packages.
+In these situations it may then be possible to "just drop a script tag" in a few
+locations to get `--no-modules` working with NPM packages. It's unclear how
+viable this is, though.
+
+# Rationale and Alternatives
+[alternatives]: #rationale-and-alternatives
+
+When developing this RFC, some guiding values for its design have been
+articulated:
```
fitzgen, line 153: Thank you for summarizing all of this in one place.

fitzgen (commented):
Thanks for this RFC, Alex. It provides answers for all of the open questions I had about the last iteration.

I have a few comments below, but these are nitpicks that I think we can largely address in implementation, and which don't need to block this RFC moving forward.

text/008-npm-dependencies.md
alexcrichton: Actually now that I think about it I think we can get rustc/cargo to learn about this dependency via `include_str!`. We could probably emit a dummy constant that doesn't end up getting used to force it to get included. That way the proc macro could actually work with the contents rather than just the path (and could put the entire contents in the custom section)

alexcrichton:
I'd like to propose that we enter the Final Comment Period for this RFC. The [original RFC](https://github.com/rustwasm/rfcs/pull/4) has been around for quite some time and has baked for quite awhile, and I'd like to help spur on final discussion for this one before implementing!

Disposition: **merge**

* [x] @fitzgen
* [x] @ashleygwilliams 
* [x] @alexcrichton

text/008-npm-dependencies.md
```diff
@@ -0,0 +1,218 @@
...
+
+Adding an NPM dependency to a Rust project will look very similar to adding an
+NPM dependency to a normal JS project. First the dependency, and its version
+requirement, need to be declare. This RFC proposes doing this in a
+`package.json` file adjacent to the crate's `Cargo.toml` file:
+
+```json
+ {
+  "dependencies": {
+    "foo": "1.0.1"
```
Pauan: This should use `^1.0.1`. I don't think we should be encouraging locking down a specific version (which is unidiomatic).

text/008-npm-dependencies.md
```diff
@@ -0,0 +1,218 @@
...
+  }
+}
+```
+
+The `package.json` file will initially be a subset of NPM's `package.json`,
```
Pauan, line 51: Are there any benefits to this restriction? Especially since we *will* support other fields in the future.

Pauan (approved):
I think this seems very reasonable for an MVP.

text/008-npm-dependencies.md
alexcrichton: I'm mostly thinking that we want to start out conservative. In the end wasm-bindgen will be producing a "merged" package.json, but that means that we in theory need a merging function for all fields of package.json. For now it's easy enough for us to define how to merge `dependencies`, but beyond that I figure we'd want to make explicit decisions about how to merge other fields rather than having something accidentally ad-hoc-ly define how to do it for us

Pauan: Could we somehow error for transitive crates but not the current crate?

Since the current crate might be published to npm, so it might want things like `name`, `version`, etc.

But transitive crates aren't intended for publishing to npm, so it's okay to restrict them to `dependencies` only.

alexcrichton: It's possible I think! My hope though is that only allowing `dependencies` is a useful-enough MVP to build on a feature like this in the future. Do you think, though, that we need to allow other fields (in some possible form) in the initial pass to be useful enough though?

Pauan: I suspect that at least `version` will be needed, since the npm version can be different from the Rust version (e.g. breaking changes in `.rs` but not in compiled `.wasm`).

Maybe not needed for the MVP though, so I don't think it should block anything. I think this RFC is fine as-is.

alexcrichton:
ping @ashleygwilliams, have you had a chance to take a look at this and review the [proposal for FCP](https://github.com/rustwasm/rfcs/pull/8#issuecomment-464214371) in the past two weeks?

alexcrichton:
I've got a sample implementation of the currently proposed version of the RFC at https://github.com/rustwasm/wasm-bindgen/pull/1305

alexcrichton:
@ashleygwilliams another ping since last week to check [the FCP proposal](https://github.com/rustwasm/rfcs/pull/8#issuecomment-464214371)

alexcrichton:
@ashleygwilliams just another ping to check out [the FCP proposal](https://github.com/rustwasm/rfcs/pull/8#issuecomment-464214371), would be great if we could get this in the near future!

chinedufn:
@alexcrichton not to speak for Ashley but I noticed from twitter that she's recently moved cross country and changed jobs  so I would imagine that she's super busy.

Not my intention to speak for her - just letting you know in case you don't already!

(@ashleygwilliams I'm **really** not trying to speak for you and just trying to call out that you might be a bit busy right now so feel free to let me know if I'm totally wrong!!!)

alexcrichton:
@chinedufn true! I'm trying to only ping at most once a week though because I want to make sure this doesn't fall off the radar but it also isn't too overwhelming. We currently require a full-team signoff (although that may change in https://github.com/rustwasm/rfcs/pull/9), so we're unfortunately stuck leaving out this feature (which I personally feel is very important!) until @ashleygwilliams checks her box.

chinedufn:
Sounds good - hadn't seen #9 - thanks!

alexcrichton:
🔔 🔔 🔔 

trivigy:
@alexcrichton The restriction just broke the builds! As part of the workflow `scripts` are included in package.json. In particular, when using webpack to do extra automation for scss, css, and a bunch of other bundling as well as prettier, and storybooks, and more. To execute on all of them and more there is a need for something like this:

```
{
  "scripts": {
    "clean": "rm -rf ./pkg ./target ./dist ./node_modules",
    "build": "webpack",
    "serve": "webpack-dev-server",
    "storybook": "start-storybook -p 6006",
    "format": "prettier --write \"./src/**/*.{js,scss}\""
  },
  "devDependencies": {
    "@babel/core": "^7.7.4",
    "@babel/plugin-syntax-dynamic-import": "^7.7.4",
    "@storybook/addon-actions": "^5.2.8",
    "@storybook/addon-knobs": "^5.2.8",
    "@storybook/addons": "^5.2.8",
    "@storybook/html": "^5.2.8",
    "@wasm-tool/wasm-pack-plugin": "^1.0.1",
    "babel-loader": "^8.0.6",
    "babel-plugin-macros": "^2.7.1",
    "clean-webpack-plugin": "^3.0.0",
    "copy-webpack-plugin": "^5.0.4",
    "css-loader": "^3.2.0",
    "fibers": "^4.0.2",
    "html-webpack-plugin": "^3.2.0",
    "paths.macro": "^2.0.2",
    "prettier": "^1.19.1",
    "react": "^16.12.0",
    "react-dom": "^16.12.0",
    "sass": "^1.23.1",
    "sass-loader": "^8.0.0",
    "style-loader": "^1.0.0",
    "ulid": "^2.3.0",
    "webpack": "^4.29.4",
    "webpack-cli": "^3.1.1",
    "webpack-dev-server": "^3.1.0"
  }
  }
}

```

In addition. Please note, that the dependencies these tools require are "devDependencies" only. I wouldn't want them to be included in the "dependencies" key in the first place. Also important to note here the use of `@wasm-tool/wasm-pack-plugin`. This one is a must because it is the only way to integrate into the storybook custom build process. They support an additional `webpack.config.js` file placed inside of `.storybook` directory.


Not entirely sure how I could work around such limitation of "only dependencies". Would appreciate some input.

trivigy:
On a separate note, this RFC is extremely nuanced because the only reason this issue surfaced in my specific use case was due to the use of `#[wasm_bindgen(module = "foo")]` module dependency. Without in, there is no lookup and merging of the package.json and therefore this change was never noticed until now.

alexcrichton:
Sorry for the difficulties you're having @trivigy! I think that unfortunately there's not a great answer today, but brainstorming on how to solve this would be greatly appreciated!

trivigy:
@alexcrichton I ended up spending the time rewriting everything in terms of `inline_js`. That's my work around for now. But I bet this exact issue will come up for others as well. Thanks for the hard work you put in and for responding.

DarthGandalf:
This broke me too. I use `npm install` to install npm dependencies, followed by `npm run build` which uses webpack which invokes wasm-pack via `@wasm-tool/wasm-pack-plugin`. Can there be an option for wasm-pack to allow any keys in `package.json` and not require user to remove all keys from it which would break other tools which read it?

There are lots of tools which use `package.json` other than the `dependencies` key:
* [`npm run`](https://docs.npmjs.com/cli/run-script)
* [Babel](https://babeljs.io/docs/en/configuration#packagejson)
* [ESLint](https://eslint.org/docs/user-guide/configuring)
* [browserslist](https://github.com/browserslist/browserslist)
* etc

While some of them have alternative ways to configure them, breaking a common practice of using `package.json` for the sake of better integration with NPM ecosystem is self-contradicting.

Please remove this restriction and simply ignore unknown keys.

alexcrichton:
@DarthGandalf at this point it's probably best to open an issue and/or PR on `wasm-bindgen`, and please feel free to do so!

Pauan:
I've given some thought to this, and I agree that we should allow for all keys in `package.json`. This makes it possible to use 1 `package.json` file (rather than 2), and it also matches what npm itself does.

PR #10, opened 2019-03-08
Add `watch` and `serve` subcommands to `wasm-pack`

fitzgen:
[**Rendered**](https://github.com/fitzgen/rfcs-1/blob/wasm-pack-watch-and-serve/text/010-wasm-pack-watch-and-serve.md)

----------------------------

cc @ashleygwilliams @alexcrichton @DebugSteven @drager

### DIFF
--- a/text/010-wasm-pack-watch-and-serve.md
+++ b/text/010-wasm-pack-watch-and-serve.md
@@ -85,9 +85,10 @@ ARGS:
 ### Implementation
 
 We can use [the `notify` crate][notify] to watch the filesystem for
-changes. Initially, we will just watch the crate directory for
-changes. Eventually, we can use [`cargo build --build-plan`][build-plan] to get
-a list of files that we should be watching.
+changes. Initially, we will use some heuristics to determine which files an
+directories to watch (see unresolved questions). Eventually, we can use [`cargo
+build --build-plan`][build-plan] to get a list of files that we should be
+watching.
 
 [notify]: https://crates.io/crates/notify
 [build-plan]: https://github.com/rust-lang/cargo/issues/5579
@@ -224,6 +225,20 @@ and add these `wasm-pack` features?
   writing Web servers, which shouldn't be *too* hard given our limited need for
   features. But ideally we shouldn't need to do this given that `miniserve`
   seems to fulfill all of our requirements.
+
+* Until we can use `cargo build --build-plan`, what heuristics should we employ
+  to guess what files and directories to watch for changes? This decision seems
+  like something that doesn't have to be written in stone in this RFC, and can
+  be experimented with during implementation and via user feedback. That said,
+  there have been suggestions for all of
+
+  * the crate's directory and its contained files and directories,
+  * walking up to find the `.git` folder at the root of a git repository and
+    watching the whole repository,
+  * finding the root of the workspace (either via `cargo-metadata` or by peeking
+    at the set of dependencies in the crate's `Cargo.toml`) and watching the
+    whole workspace,
+  * and some combination of these options.
 
 [https]: https://crates.io/crates/https
 [miniserve]: https://crates.io/crates/miniserve
--- a/text/010-wasm-pack-watch-and-serve.md
+++ b/text/010-wasm-pack-watch-and-serve.md
@@ -107,9 +107,13 @@ defaults to the current directory. It has the usual flags for controlling the
 build profile, and the target environment. It takes extra options after `--`
 that get passed through straight to `cargo build`.
 
-Additionally, it has a `--no-watch` flag to disable watching the crate for
-changes to kick off automatic re-builds, and a `--port` option to specifiy the
-port the local server should bind to.
+Additionally, it has
+
+* a `--no-watch` flag to disable watching the crate for changes to kick off
+  automatic re-builds,
+* a `--port <port>` option to specifiy the port the local server should bind to,
+* and a `--root <root>` option to specify the root directory to serve files
+  from.
 
 ```
 USAGE:
@@ -132,6 +136,7 @@ OPTIONS:
     -t, --target <target>      Sets the target environment. [possible values: browser, nodejs, no-modules] [default:
                                browser]
     -p, --port <port>          Bind to this port number with the local development HTTP server. [default: 8000]
+    -r, --root <root>          Use this directory as the root to serve files from. [default: crate directory]
 
 ARGS:
     <path>                The path to the Rust crate.
--- a/text/010-wasm-pack-watch-and-serve.md
+++ b/text/010-wasm-pack-watch-and-serve.md
@@ -1,0 +1,230 @@
+- Start Date: 2019-03-06
+- RFC PR: https://github.com/rustwasm/rfcs/pull/10
+- Tracking Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Add two new sub-commands to `wasm-pack`:
+
+1. `watch`: watch the crate's files and automatically rebuild whenever they
+   change.
+2. `serve`: start an HTTP server for the crate directory on `localhost`.
+
+# Motivation
+[motivation]: #motivation
+
+Enable a smooth and complete local development experience for users who are not
+using a JavaScript bundler.
+
+In particular, we would like to [remove the `npm` and bundler usage from our
+Game of Life tutorial][no-bundler-in-docs] without requiring readers to install
+any additional tools other than the Rust toolchain and `wasm-pack` when setting
+up their development environment. The goal here being less moving parts to learn
+and tools to wrangle when first exploring Rust and WebAssembly.
+
+Additionally, we would like to make sure that the local development server uses
+the `application/wasm` MIME type when serving Wasm binaries. This enables the
+`WebAssembly.instantiateStreaming` fast-path for users.
+
+[no-bundler-in-docs]: https://github.com/rustwasm/book/issues/150
+
+# Stakeholders
+[stakeholders]: #stakeholders
+
+The stakeholders are primarily people that are doing Rust and Wasm development
+without a JavaScript bundler, which potentially will include all of our new
+users who are just starting the Game of Life tutorial. Most of this demographic
+is probably not watching the RFCs repository, and only some are regularly coming
+to working group meetings, so it makes sense to advertise this RFC in TWiRaWA
+and on our @rustwasm Twitter account.
+
+# Detailed Explanation
+[detailed-explanation]: #detailed-explanation
+
+We will add two new subcommands to `wasm-pack`:
+
+1. `watch`
+2. `serve`
+
+## `wasm-pack watch`
+
+The `watch` subcommand will observe the crate directory for changes and
+automatically re-build when files change.
+
+### Command Line Interface
+
+The `watch` command takes an identical set of command line arguments as
+`wasm-pack build` does.
+
+```
+USAGE:
+    wasm-pack watch [FLAGS] [OPTIONS] [path] [-- <extra_options>...]
+
+FLAGS:
+        --dev              Create a development build. Enable debug info, and disable optimizations.
+        --no-typescript    By default a *.d.ts file is generated for the generated JS file, but this flag will disable
+                           generating this TypeScript file.
+    -h, --help             Prints help information
+        --profiling        Create a profiling build. Enable optimizations and debug info.
+        --release          Create a release build. Enable optimizations and disable debug info.
+    -V, --version          Prints version information
+
+OPTIONS:
+    -m, --mode <mode>          Sets steps to be run. [possible values: no-install, normal, force] [default: normal]
+    -d, --out-dir <out_dir>    Sets the output directory with a relative path. [default: pkg]
+    -s, --scope <scope>        The npm scope to use in package.json, if any.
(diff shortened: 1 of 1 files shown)


text/010-wasm-pack-watch-and-serve.md
```diff
@@ -0,0 +1,230 @@
...
+
+ARGS:
+    <path>                The path to the Rust crate.
+    <extra_options>...    List of extra options to pass to `cargo build`
+```
+
+### Implementation
+
+We can use [the `notify` crate][notify] to watch the filesystem for
+changes. Initially, we will just watch the crate directory for
```
alexcrichton: One thing I might recommend is that initially instead of watching just the crate directory we have a heuristic which attempts to find a git repository which contains the crate going up the tree. If found we watch the entire git repository, and if not found we watch the entire crate directory.

All of cargo/wasm-pack/wasm-bindgen are quite speedy when a change was made that doesn't actually change the output artifact, so I think it'd be fine to be a bit more coarse here in terms of changes that we listen for.

Now that being said this is something that can be tweaked at any time, so no need to hold up the RFC on a point like this!

text/010-wasm-pack-watch-and-serve.md
```diff
@@ -0,0 +1,230 @@
...
+
+### Implementation
+
+Rather than integrating an HTTP server into the `wasm-pack` binary itself, we
+should leverage its ability to download and run other binaries. Exactly *which*
+local HTTP server binary is left as an unresolved question.
+
+The subcommand will start the local HTTP server and then execute the `watch`
+subcommand (unless `--no-watch` is supplied).
+
```
alexcrichton, line 154: What do you think about adding a `--root` option to the `wasm-pack serve` subcommand? The HTTP server will presumably need to figure out a filesystem root path to start hosting static files from, and while not specified here the cwd seems like a reasonable default, but having a `--root` option as well to configure it may be reasonable. For example if you customize `--out-dir` you may want to also customize `--root`, like:

```
wasm-pack serve --out-dir html/dist --root html
```

(or something like that)

text/010-wasm-pack-watch-and-serve.md
```diff
@@ -0,0 +1,230 @@
...
+        --profiling        Create a profiling build. Enable optimizations and debug info.
+        --release          Create a release build. Enable optimizations and disable debug info.
+    -V, --version          Prints version information
+        --no-watch         Do not watch the crate for changes to automaticaly rebuild.
+
+OPTIONS:
+    -m, --mode <mode>          Sets steps to be run. [possible values: no-install, normal, force] [default: normal]
+    -d, --out-dir <out_dir>    Sets the output directory with a relative path. [default: pkg]
+    -s, --scope <scope>        The npm scope to use in package.json, if any.
+    -t, --target <target>      Sets the target environment. [possible values: browser, nodejs, no-modules] [default:
```
alexcrichton, line 136: An interesting aspect of the `serve` command is that I think it basically doesn't work if you either don't pass `-t` or if you pass an incompatible value like `-t nodejs`. 

I wonder if we should have some stricter validation (and documentation) which says that `wasm-pack serve` is only compatible with the `no-modules` and `web` (new) target?

text/010-wasm-pack-watch-and-serve.md
```diff
@@ -0,0 +1,230 @@
...
+hyper, tide, etc:
+
+* **Pros:**
+  * The `wasm-pack serve` command is ready to roll immediately after `wasm-pack`
+    is installed, and doesn't need to hit the network the first time you run
+    it. This situation only arises when folks who already downloaded `wasm-pack`
+    do their first build/serve offline, disconnected from the internet. This
+    seems like a niche enough situation that we can take the trade off.
+* **Cons:**
+  * Building the HTTP server into the `wasm-pack` binary is less robust: by
```
alexcrichton, line 175: FWIW this doesn't seem like too much of a con to me, if we were dealing with C/C++ I'd definitely agree but with Rust error recovery is pretty simple in-process. 

That being said I still agree the next con outweighs the pro

text/010-wasm-pack-watch-and-serve.md
```diff
@@ -0,0 +1,230 @@
...
+    shelling out to an external binary for our local server, we get process
+    isolation, which makes error recovery simpler.
+  * More work to implement. `wasm-pack` already has plenty of infrastructure for
+    downloading and working with external binaries, so using external HTTP
+    servers should be fairly easy to implement.
+
```
alexcrichton, line 181: Thinking about this, one feature of `webpack-dev-server` is that when you request a page it actually waits to send any data until the current build is done. In that sense I think that a con of using an external binary is that we can't implement logic like that.

For example if I make a change and then quickly go and refresh a web page, I might get stale output if the build hasn't finished yet.

I wonder though if we could perhaps destroy `--out-dir` on each build to ensure that if the page is loaded stale artifacts are never served up? (not a great experience but perhaps better than seeing stale data)

alexcrichton (commented):
Thanks for writing this up @fitzgen! I think this wasn't something I realized I wanted until you articulated it...

Long-term I'd love to explore possibilities of removing the static file server in `wasm-bindgen-test-runner` as well. If we could leverage `wasm-pack`'s own internal server somehow that'd be great. I think that's pretty squarely outside the scope of this RFC, though, and we can deal with that later.

text/010-wasm-pack-watch-and-serve.md
```diff
@@ -0,0 +1,230 @@
...
+### Command Line Interface
+
+The `serve` subcommand takes a superset of the arguments that `wasm-pack build`
+takes. Like other `wasm-pack` subcommands, it takes an optional path to the
+crate directory as a positional argument, but if that is missing, then it
+defaults to the current directory. It has the usual flags for controlling the
+build profile, and the target environment. It takes extra options after `--`
+that get passed through straight to `cargo build`.
+
+Additionally, it has a `--no-watch` flag to disable watching the crate for
```
drager: Maybe this should also be able to be controlled from wasm-pack's toml configuration? https://rustwasm.github.io/wasm-pack/book/cargo-toml-configuration.html

drager:
I think this is a great idea! Both commands and functionality makes sense to implement. I agree that if wasm-pack implements the `serve` command, watching and automatic rebuild is required together with serve.

text/010-wasm-pack-watch-and-serve.md
```diff
@@ -0,0 +1,230 @@
...
+[detailed-explanation]: #detailed-explanation
+
+We will add two new subcommands to `wasm-pack`:
+
+1. `watch`
+2. `serve`
+
+## `wasm-pack watch`
+
+The `watch` subcommand will observe the crate directory for changes and
```
chinedufn, line 52: Perhaps not in v1...? - but I think it would be help to also take a look at the Cargo.toml, look at the dependencie’s and also watch any path dependencies that we see.

Most of my non trivial projects are in a workspace with different path deps - so just watching one crate wouldn’t be of use to me.

I see that there are plans to make use of the build-plan so that would work too. Not familiar enough to know which is an easier approach.

That being said - this seems like a nice step in the right direction - just sharing feedback!

text/010-wasm-pack-watch-and-serve.md
fitzgen: I wasn't thinking just memory, but also other resources like sockets, but yeah much less of a deal in Rust.

fitzgen:
Just pushed two more commits ådding an unresolved question about how to choose which files to watch and added a `--root <root>` option to the proposed `wasm-pack serve` subcommand.

jrvidal:
Random thought out of the blue: there is a related, potential use case not covered by this RFC which is **integrating with a JS bundler watch process** (already hinted by @chinedufn I think)

In a big JS project where Rust is introduced incrementally, it's very likely that there is already a bundler in place (with a sizable configuration) taking care of the watching and rebuilding. There currently exist a few `<bundler>-rust-loader-plugin-thingy` for the popular bundlers but, from my superficial impressions, they focus in _raw_ rustc usage and single Rust files exposing the barebones Wasm FFI.

Now, if `wasm-pack` wanted to help with this use case, I suspect it would have to yield the `watch`ing, and instead expose as much build-plan information as possible so the loaders can kick this information up to the bundler's own watcher.

OTOH, it's also possible to write loaders top of `cargo` alone and let `wasm-pack` completely out of it. But I wonder if there is some functionality that will have to be re-implemented by the loader. I'm thinking about the tiny things that `wasm-pack` does to ensure a smooth experience, like checking your Rust installation, or downloading `wasm-bindgen` binaries (these would be the extra things that would be exposed in the "build plan").

Of course this might very well be out of scope for an initial MVP, or for `wasm-pack` outright :sweat_smile:

fitzgen:
> **integrating with a JS bundler watch process**

I think we can split this scenario into two designs:

1. The JS bundler asks `wasm-pack` to watch for changes, since `wasm-pack` knows more about the Rust sources than the JS bundler does. Then `wasm-pack watch` is exactly what is needed, and the JS bundler can watch the `pkg` directory to get its downstream updates. The potential downside is overlapping watch directories with the project's other JS, and in an ideal world a single file would only ever be watched by a single thing.

2. The JS bundler wants to watch for changes and then tell `wasm-pack` when to rebuild. This can potentially solve the "only a single watcher for a single file" thing. In this scenario, the JS bundler wants `cargo build --build-plan` to figure out which files to watch, same as we want for `wasm-pack watch`. It can re-invoke `wasm-pack build` on each change however it sees fit. The latency of `wasm-pack build` is pretty good too, so false positives shouldn't be a big deal, and means that until `cargo build --build-plan` exists, it is free to use its own heuristics for choosing what to watch, similar to what we plan on shipping for `wasm-pack watch` initially.

Most importantly, in both integrating-with-JS-bundler-watching scenarios, I don't see anything that needs to change in this RFC, or any fundamental design contradictions.

fitzgen:
The discussion here seems to have slowed down a bit, and there seems to be broad consensus that we want this general feature, and that we can continue to improve and evolve the subcommands during and after initial implementation.

Therefore, I'd like to propose that this RFC enter its **final comment period** with disposition to **merge**.

@rustwasm/core team members to sign off:

* [x] @alexcrichton 
* [ ] @ashleygwilliams 
* [x] @fitzgen

alexcrichton:
Still quite a fan of this feature, so checking my box!

ashleygwilliams:
Hey! So ~as I said previously~ I'd like to postpone this until we've shipped 0.8.0 (April 26th). At that point I'll be in a better position to evaluate and respond to this RFC (because we will be closer to being able to prioritize these features and I'd prefer to not have a large amount of time between the design and implementation work.)

---
EDIT: I'm generally super in favor of this, and just anticipate I may have some design input, not that I would reject these features.

alexcrichton:
Wanted to ping on this! @ashleygwilliams and @drager do y'all feel that wasm-pack is in a position where this can be reviewed/implemented?

drager:
@alexcrichton I'm not sure in which sprint we can do this in wasm-pack at the moment. Both @ashleygwilliams and I have been pretty busy these last two weeks and we have a couple of issues left labeled [sprint](https://github.com/rustwasm/wasm-pack/issues?q=is%3Aissue+is%3Aopen+label%3Asprint) and some new issues I think we need to deal with and/or investigate.

Personally, if we can fit some new functionality into the next sprint I think this should probably be one of those features.

alexcrichton:
Note though that this isn't really about implementation work, for now this is purely just "needs a review and discussion on the RFC" style work.

ashleygwilliams:
I'm on vacation but I 100% think we should open this up to conversation about design as we should be able to do this, likely not the next sprint (as originally intended) but the one after.

boringcactus:
is there anything community members can do to help finalize this RFC?

if we need more opinions about embedded vs external http handling, my opinion is that https://github.com/rustwasm/rfcs/pull/10#discussion_r263795579 is a good reason to build the server directly into wasm-pack instead of shelling out to an external server. having weird timing conditions where users can sometimes get stale data, or nothing at all, substantially weakens the user experience, which is a serious drawback in subcommands that are designed to streamline the user experience.

boringcactus:
Also worth noting is that per https://github.com/rust-lang/cargo/issues/7614 `cargo build --build-plan` is being removed, so `watch` will have to decide on its own which files should be watched.

boringcactus:
I've thrown together a draft of a plausible implementation at https://github.com/rustwasm/wasm-pack/pull/745, partially out of curiosity and partially because @ashleygwilliams suggested that having a short turnaround time between design and implementation would be helpful.

My implementation uses an integrated server and a `RwLock` to pause the server while a build is running, and adds less than 200 lines of code, so it seems that an integrated server would not in fact be more work to implement.

jasonwilliams:
What's the blocker for this RFC?

pastelmind:
> The JS bundler wants to watch for changes and then tell wasm-pack when to rebuild.

This didn't work for me (using Rollup) when I integrated webpack into my build config, because it created an infinite loop. When wasm-pack rebuilt the Rust code, the regenerated `pkg/` dir caused Rollup to rebuild, which invoked wasm-pack, which regenerated `pkg/`, which triggered Rollup, and so on.

I ended up running a separate watcher that only watches Rust code and invokes wasm-pack.

jasonwilliams:
In the meantime if anyone is using esbuild https://github.com/Tschrock/esbuild-plugin-wasm-pack supports watching Rust files (as well as any JS files)

PR #11, opened 2019-08-16
Enable exported Rust structs to specify prototypal inheritance

### LINKED ISSUE #210: Add ability to import and extend JS classes
JS classes are heavily used building block in UI development.

It's common for a library like ReactJS or Polymer to provide a component class to extend from.

Thus, there should be an easy way to interoperate with classes, extend them in Rust, and export them back to JS.

I don't know about Rust to know what's possible yet. But I imagine we could use attributes to annotate when a Rust struct inherits from a JS class?

eggyal:
[**Rendered**](https://github.com/eggyal/rfcs/blob/master/text/000-prototypal-inheritance.md)

Fixes https://github.com/rustwasm/wasm-bindgen/issues/210 and https://github.com/rustwasm/wasm-bindgen/issues/1721.

This is an early draft, but I'd be really grateful for feedback at this stage—not sure whether I'm detailing it correctly, nor where exactly to go with some of the blank sections?

### DIFF
--- a/text/000-prototypal-inheritance.md
+++ b/text/000-prototypal-inheritance.md
@@ -39,6 +39,20 @@ In respect of the latter group, various related issues have been identified in t
 # Detailed Explanation
 [detailed-explanation]: #detailed-explanation
 
+### Overview
+
+At present, values of exported Rust types are owned and moved around by the user like any other value.  If passed "by value" to JavaScript, they are moved into a boxed `WasmRefCell` the address of which is opaquely sent over the ABI; that address is held in the `ptr` property of an ephemeral shim object, such property being returned over the ABI (and cleared) if/when the value is passed "by value" back to Rust.
+
+This RFC proposes a fundamental change from that status-quo.  Values of exported Rust types will now remain permanently in a boxed `WasmRefCell` and cannot be owned or moved around by the user: this enables the JavaScript shim to exist (and properly function) for the object's entire lifetime, which may be required by virtue of side-effects from calling the object's super-constructors (e.g. if the object is registered with some form of global registry, as is the case with Web Components).
+
+Consequently, users in Rust can only ever have a `Ref` or `RefMut` that derefs to their object.  Ownership of the object is effectively controlled by the long-lived JavaScript shim, being transferred back to Rust only for the value to be dropped when it (and the shim) are no longer required.
+
+Furthermore, this RFC proposes that exported Rust types are themselves modified to "own" the Rust values for each object on their prototype chain through the injection of an additional field.  This provides users with native access to inherited methods and properties (irrespective of whether the prototype is native to Rust or JavaScript).
+
+A little additional bookkeeping is required in order for methods invoked on the JavaScript shim objects to have the correct receiver in Rust.
+
+### Details
+
 1. Exported type `T` **MAY** be annotated with an additional input to its `#[wasm_bindgen]` outer attribute: namely `prototype=path::to::Parent`, where `Parent` is either an imported or another exported wasm-bindgen type.
 
 2. A field is injected into the definition of `T`; it is either named `__proto__` or appended to the existing unnamed field list, as appropriate (unit types are converted to having a single field named `__proto__`).  This injected field is either of type `Parent` (if the `prototype` input was specified above), or else of type `JsValue` (otherwise).  Users never need access this field directly; instead `T` implements both `Deref<Target=Parent>` (or `Target=JsValue`, as appropriate) and `DerefMut` that return references to it.
--- a/text/000-prototypal-inheritance.md
+++ b/text/000-prototypal-inheritance.md
@@ -1,0 +1,315 @@
+- Start Date: 2019-08-16
+- RFC PR: 
+- Tracking Issue: 
+
+# Summary
+[summary]: #summary
+
+Enable each exported Rust struct to specify the prototype chain acquired by instances of its generated JavaScript shim.
+
+This inheritance relationship should also be reflected on the Rust side, in the same way that `#[wasm_bindgen(extends)]` effects such relationships in classes imported from JavaScript by implementing relevant traits on the associated Rust struct.
+
+
+# Motivation
+[motivation]: #motivation
+
+Some JavaScript APIs expect their clients to provide objects that inherit behaviour from another (usually supplied by the API) by ensuring that the latter is on the former's prototype chain.  For example:
+
+* the object referenced by the `prototype` property of [`React.Component`](https://reactjs.org/docs/react-component.html) is expected to be on the prototype chain of one's [stateful React components](https://reactjs.org/docs/state-and-lifecycle.html); and
+
+*  the object referenced by the `prototype` property of the relevant DOM constructor is expected to be on the prototype chain of one's [custom elements](https://html.spec.whatwg.org/multipage/custom-elements.html).
+
+Correct consumption of such APIs in Rust via `wasm-bindgen` requires each object returned to JavaScript to have the appropriate prototype chain.
+
+
+# Stakeholders
+[stakeholders]: #stakeholders
+
+The major stakeholders in this RFC are:  
+
+* JavaScript library developers whose APIs expect clients to use prototypal delegation; and
+
+* Users of `#[wasm_bindgen]` who wish to consume such APIs.
+
+In respect of the former group, we have already identified the developers of React and the authors of the Web Component specification. ***It is not yet known how feedback can best be solicited from this group—input welcome!***
+
+In respect of the latter group, various related issues have been identified in the `rustwasm` repositories and comments will be posted thereto directing contributors to this RFC.
+
+
+# Detailed Explanation
+[detailed-explanation]: #detailed-explanation
+
+There are three key problems with the status quo that currently prevent exported types from participating in JavaScript's prototypal inheritance:
+
+1. There is currently no way in Rust to specify the inheritance relationship and thereby guide what [`extends`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/extends) clause should be generated on the JavaScript shim class, which makes it more difficult (if not impossible) to construct objects in the manner required by many APIs; indeed, Web Components' customized built-in elements explicitly require use of the `extends` clause, per the normative [core concepts](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements-core-concepts) section of their specification.
+
+2. Even if an `extends` clause is somehow generated, the shim's `constructor()` (which presently just wraps whatever WASM function was annotated with `#[wasm_bindgen(constructor)`) **must** invoke its super-constructor before returning: but there is presently no way to make this invocation from within WASM.  Moreover, the shim's `constructor()` presently returns an object other than `this`, which violates the expectations of many APIs (including the explicit [conformance requirements of Web Components' custom elements](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-element-conformance)); changing this behaviour could necessitate the delegated WASM constructor having access to `this` in order to perform correct initialisation (e.g. where there is some dependency upon inherited behaviour or state set by super-constructors).
+
+3. `wasm-bindgen`'s existing approach to passing WASM objects to JavaScript does not provide for their polymorphism into parent types on their return to WASM.  In particular, when WASM transfers out to JavaScript an opaque pointer (in fact the address of a boxed `WasmRefCell`) to some `Child` instance and later receives that pointer back as part of an incoming FFI that expects a pointer to a `Parent`, any resulting attempt to dereference the pointer as a `Parent` when it is in fact a `Child` will of course be Undefined Behaviour—despite the fact that this *should* be permitted under the inheritance relationship.
+
+This RFC proposes a solution to all of these points, and furthermore also proposes a solution to the related problem of obtaining a [`JsValue`](https://rustwasm.github.io/wasm-bindgen/api/wasm_bindgen/struct.JsValue.html) for any given export or import which could be used, for example, to pass a `struct` to a JS API that expects a constructor function (such as [`web_sys::CustomElementRegistry::define`](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.CustomElementRegistry.html#method.define)).
+
+These problems are addressed in this order below, but consideration is at first only given to how exported types are instantiated from JavaScript; the problem of correctly instantiating exported types from within WASM is not visited until the third section (on FFI polymorphism).
+
+### Generating `extends` clause in shim classes
+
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for writing this up! This is definitely something I think we need to support in one form or another, and it's great to see progress on this!

One of my main questions would be how we're going to implement this all in terms of generated code. It seems to me that when you specify a superclass you typically want to, in methods, be able to access the `JsValue` for the superclass (although probably typed instead of bland) as well as the custom `struct` data fields (what you defined in `Rust`). That doesn't map super well to how the bindings are generated today where an exported Rust struct ends up having a JS value on the other side but Rust can't really ever access it.

Would you be able to expand the detailed design with snippets and examples of what the generated code would look like? (both on the Rust and JS side of things)

eggyal:
@alexcrichton, if I've understood correctly, then it seems to me that there are two possibilities:

1. The superclass is an `extern type`, imported from JavaScript—in which case, it is already heap-allocated and its `JsValue` is accessible through the wrapping `struct`; or

2. The superclass is a natively-defined Rust-type that is exported to JavaScript via manual `#[wasm_bindgen]` attributes (as with any other exported type)—in which case, there is no need it to have a `JsValue`: the JS that is already generated defines a class that will suffice perfectly for use in the subclass's `extends` clause?

Perhaps I've misunderstood your point?

eggyal:
@alexcrichton, having re-read your comment I realise now that I had misunderstood.

First of all, I hadn't seen it being necessary to retrieve the superclass's `JsValue` at runtime—at least not for the use cases I had in mind.  Can you elaborate on why you think this might be required?

Second, should it be necessary then I suppose there could be a Rust trait `Subclass` that defines a static `fn superclass() -> JsValue`, implementations of which are generated for every exported Rust struct.  For imported superclasses this should be easy enough (as the superclass will already be on the heap as mentioned in my previous comment); but I see that something would need to be done for Rust-native superclasses—presumably they would also have to be allocated on the heap after export?

eggyal:
@alexcrichton: Apologies that I've been around the houses before arriving here, but having now delved under the hood a fair bit I think I'm beginning to gain a better appreciation for the issues.

It seems to me that the inheritance relationship must be encoded within the custom `__wasm_bindgen_unstable` section of the compiled `.wasm`: in particular, as part of the encoding of each `Struct`.  Since the relevant parent will be defined elsewhere within that same section of that same `.wasm`, as either (a) an `ImportType` or (b) another `Struct` (respectively for the two cases detailed above), ideally one would somehow internally reference (rather than duplicate) that definition.

AIUI, each use of `#[wasm_bindgen]` results in (amongst other things) a static with the `#[link_section = "__wasm_bindgen_unstable"]` attribute, such static containing an encoding of the bindings required by the annotated code.  `rustc` provides no guarantee of the order that these encoded bindings will appear in the custom section, and therefore referencing by their position is impossible.  Referencing by name is also troublesome, since there is no guarantee of uniqueness.  I therefore believe that the only solution may be to annotate each binding with a unique identifier in order that such identifier can be used by referencing bindings.  Do you agree?  If so, do you think this may be worthy of an RFC in itself?

As regards implementation, obtaining the identifier of a referenced binding will require name resolution (not available during macro expansion); I therefore propose that the static that contains the encoded binding for the custom section be initialised in part from calling a `const fn` on the referenced type (rather than purely an expanded literal as is currently the case).   For example:

```rust
#[wasm_bindgen]
struct Parent();

#[wasm_bindgen(superclass=Parent)]
struct Child();
```

would expand to something along the lines of the following (amongst other things):

```rust
trait WasmBindgenReferencable {
    fn wasm_bindgen_id() -> u32;
}

struct Parent();
impl WasmBindgenReferencable for Parent {
    const fn wasm_bindgen_id() -> u32 { 0 }
}

#[link_section = "__wasm_bindgen_unstable"]
pub static __WASM_BINDGEN_GENERATED_1: [u8; 123] = {
     static _INCLUDED_FILES: &[&str] = &[];

     *b"..." // includes id of 0 in binding of Parent
}

struct Child();
impl WasmBindgenReferencable for Parent {
    const fn wasm_bindgen_id() -> u32 { 1 }
}

#[link_section = "__wasm_bindgen_unstable"]
pub static __WASM_BINDGEN_GENERATED_1: [u8; 123] = {
     static _INCLUDED_FILES: &[&str] = &[];

     combine!(*b"...before...", <Parent as WasmBindgenReferencable>::wasm_bindgen_id(), *b"...after..."
}
```

alexcrichton:
Heh yeah these are some of the things I was thinking of! To be clear I haven't thought a huge amount about this feature, so I don't really have all the answers. A good driving use case will help design this, and I figured web components was one which would need access to the actual JS object to do things like modify the DOM, but I could be wrong.

I'm not entirely sure how we're going to draw all these connections between parent/child classes. IDs and byte concatenation is one (although I don't know how to do that) but we could possibly just use strings as well. In any case this is where I think it'd be helpful to prototype something that's workable and then we can have that help influence the design of the RFC and such.

eggyal:
@alexcrichton: Agreed.  I've actually been working on a prototype and have made some good progress—but currently stuck on how to allocate an internal identifier to referenceable types:

Originally I thought just to maintain a static counter within the macro implementation, which would be tokenized to a literal (similar to how `generated_static_name` is currently made unique)—but I've discovered that each crate that invokes the macro has *its own* instance of the static, and consequently collisions occur.  On top of which, incremental compilation could present problems. See [StackOverflow](https://stackoverflow.com/a/52914046) and https://github.com/rust-lang/rust/issues/44034 for discussion of crate-local persistence, but of course we actually need build-global persistence.

I've since been exploring obtaining a suitable identifier from `std::any::TypeId::of<T>()` and then (through unsafe methods) accessing its underlying data—but calling that function in const/static contexts is currently gated behind the unstable feature [`const_type_id`](https://doc.rust-lang.org/beta/unstable-book/library-features/const-type-id.html) (and I haven't yet found any way to access the underlying data without touching some other gated unstable feature, e.g. [`const_raw_ptr_deref`](https://doc.rust-lang.org/beta/unstable-book/language-features/const-raw-ptr-deref.html), [`const_fn_union`](https://doc.rust-lang.org/beta/unstable-book/language-features/const-fn-union.html) or [`const_transmute`](https://doc.rust-lang.org/beta/unstable-book/language-features/const-transmute.html)).  Since such gated features would have to be enabled in the crate(s) that use `#[wasm_bindgen]`, this does not seem to be suitable either.

Perhaps using a string is the right approach after all?

alexcrichton:
I think it's probably fine to just stick with something that works for now , this is a relatively minor portion of the feature set in this change :)

eggyal:
> I think it's probably fine to just stick with something that works for now , this is a relatively minor portion of the feature set in this change :)

Yes, but what is "something that works"?  As seen from rustwasm/wasm-bindgen#1737, generating IDs from the system clock (which works on my development machine) is not working on the CI server... struggling to find a solution here, so really grateful if anyone has any thoughts!

Meanwhile, since it's working for me locally I've started experimenting with Web Components (and in particular creating a Rust-native customized built-in element)... and now I've hit the hurdle I think @alexcrichton was alluding to in [his first comment](https://github.com/rustwasm/rfcs/pull/11#issuecomment-522712741): in order to pass one's type as the `function_constructor` argument of [`web_sys::CustomElementRegistry::define`](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.CustomElementRegistry.html#method.define), one needs to obtain a reference to it as a `js_sys::Function`.  My current thinking for this is:

1. On loading the shim JS module, add to the heap a reference to each exported object and populate a map of export names to heap indices;

2. Import into Rust from JS a lookup function into that map; and

3. Implement a trait on exported structs that provides a method calling into that lookup function in order to achieve the relevant constructor as a `&js_sys::JsFunction`.

I'll work on extending my prototype implementation to add this functionality unless there are obvious pitfalls I should avoid?

torch2424:
@eggyal This was brought to my attention at the Rust/Wasm WG meeting, and just wanted to give a few comments, hope I'm not driving the current discussion away:

1. This is super rad! 😄 Stoked that you are driving this, and I am sure it will be a huge help in a lot of rust/wasm apps 👍 🎉 Also, I think the API is very clean, I write a ton of JS, and a bit of rust/wasm, and I think it looks easy to use and makes sense.

2. The only thing I would suggest is naming? I think `superclass` makes sense in both ES6 and Rust land. But there is still a lot of popular JS libraries that don't use ES6 classes, and depend on prototypical inheritance. From your example:

```rust
// superclass imported from JS
#[wasm_bindgen]
extern "C" {
    pub type ImportedParent;
}
#[wasm_bindgen(superclass=ImportedParent)]
pub struct ChildOfImportedParent { ... }
```

If `ImportedParent` is not a class, but let's say, a popular library like [lodash](https://github.com/lodash/lodash) a user wants to extend / override with wasm based functions. Then, `superclass` doesn't make as much sense? Thus, I think renaming `superclass` to `prototype`, e.g `#[wasm_bindgen(prototype=ImportedParent)]`, makes sense. Though, I don't feel super strongly about this, as you could easily say, `prototype` doesn't make sense for rust developers.

Hope that helps, thanks! 😄 👍

Pauan:
Personally, I think the attribute should be called `extends`, since that matches the ES6 syntax, and it's also [consistent with wasm-bindgen itself](https://rustwasm.github.io/docs/wasm-bindgen/reference/attributes/on-js-imports/extends.html).

But that's all bikeshedding stuff which should be easy to change.

eggyal:
@torch2424 @Pauan I agree that `superclass` may not be the best choice, but I deliberately chose not to use `extends` because I didn't want to overload its semantics.  I'm definitely open to it, if the community is happy.  For now, I've updated the RFC to use `prototype`.

I've also fleshed out the "Detailed Explanation" section quite a bit more to reflect the work I've undertaken in https://github.com/rustwasm/wasm-bindgen/pull/1737 (together with further changes not yet pushed).  Would very much welcome critique of this design!

eggyal:
I've moved my earlier comments (to the extent they're still relevant) into the draft RFC where they now have more context and will hopefully make more sense; this also tidies up the conversation on this PR to (hopefully) facilitate wider discussion.

Pauan:
> @alexcrichton That doesn't map super well to how the bindings are generated today where an exported Rust struct ends up having a JS value on the other side but Rust can't really ever access it.

Are you referring to the JS `class` or the JS object? My understanding is that we [*can* access the JS object](https://github.com/rustwasm/wasm-bindgen/blob/master/crates/backend/src/codegen.rs#L192-L215).

eggyal:
> @Pauan Are you referring to the JS `class` or the JS object? My understanding is that we [_can_ access the JS object](https://github.com/rustwasm/wasm-bindgen/blob/master/crates/backend/src/codegen.rs#L192-L215).

I believe that @alexcrichton was referring to the `class` (apologies, I deleted the comment to which he was replying in trying to tidy this issue up a bit and move pertinent discussion into the RFC where it has better context).  A solution to this is set out in the RFC and has been implemented in the prototype.

For the JS object, you are correct that the `JsValue` can already be accessed—but this would need to change a little under the current draft RFC (see section "Object instantiation").

eggyal:
@Pauan, on reflection, it's actually a lot more difficult in an inheritance scenario.

The `impl From<#name> for JsValue` to which you referred wraps the Rust instance in its JS shim and obtains a `JsValue` representing the result: however, this "wrapping" currently uses `Object.create()` in order to avoid calling the shim constructor... this is acceptable for the status quo, where the shim object is a mere transparent proxy for the Rust instance—but in an inheritance scenario, proper instantiation of the shim may require super-constructor invocation (and indeed, it may no longer be acceptable for a single object to be represented by multiple shims over the course of its lifetime).

I therefore don't think that post-instantiation wrapping is straightforward for derived objects.  Far better, I think, to require that shims for derived objects be instantiated upon construction and live for the duration of the object's lifetime.  Of course, this does mean that the current implementation of that `From` trait is no longer appropriate.

I originally thought that a smart pointer holding the `JsValue` but that derefs in some way to the Rust object could solve this—but I'm beginning to see that it does not, since (without some sort of reverse-lookup registry) Rust instance methods would not have any means of accessing it: this is something that they will require in order to, for example, invoke methods on their prototype chain.

One idea could be to inject into exported derived objects a field that holds their prototype instance (which would be the typed `JsValue` of the shim for types that derive from imported types, or the actual Rust object for types that derive from exported types), which could then be returned by an implementation of `Deref<Target=parent_type>::deref`.  While this would force construction to always fall within our control (where we can ensure the shim is properly instantiated), it would of course break pattern matching and perhaps other things too...

Evidently this is a much much much more involved idea than I had thought at the outset!

Still working on it...

eggyal:
@alexcrichton @Pauan Following further learnings, I've completely rewritten the RFC but feel it's now extremely close to a workable solution.  I have a little more to do to fully implement this, but very grateful for any & all thoughts at this stage.

alexcrichton:
Sorry for taking awhile to get back to this @eggyal, but after reading this over I'm wondering if we can perhaps try to expose some of the lower level building blocks here? There's a lot of stylistic and library-style decisions that `wasm-bindgen` would have to make to implement this it looks like, but it might be cool if we could instead provide some building blocks to let iteration on various syntaxes happen in the ecosystem?

eggyal:
That's definitely a good idea, @alexcrichton: this proposal turned out to be significantly more complex than I (naively) anticipated at first—and whilst I have got a working implementation, I made a number of choices along the way that with more considered thought would likely have been decided differently.  Nevertheless, as you suggested early on, undertaking this implementation greatly improved my understanding of the issues...

I shall try to pull out some building blocks as you suggest.

Nachasic:
Any progress on this?

eggyal:
@Nachasic I've not looked at this for a while, but wasm-bindgen's (then?) current approach:

* instantiated shims via `Object.create()`, whereas parent classes that are native to JS may *require* that their constructors are invoked (such as is the case with custom HTML elements);

* could result in shims that are only semi-hobbled once "freed" because code that is native to JS might store state on the JS shim object itself; and

* instantiated a new shim object for the same underlying Rust struct every time it was shimmed—which may violate JS object identity assumptions.

This proposal addressed these problems by making some changes to the JS/WASM bridge; in particular, the JS shim object becomes instantiated (by calling JS constructors) upon the Rust struct's creation and is maintained until the Rust struct is dropped; furthermore the Rust struct is only accessible through an `Rc` in order to prevent it being dropped while still referenced from JS.

However this resulted in significant changes to the API, making it somewhat less ergonomic—in particular, you can no longer take ownership of exported types, and must instantiate them using a macro; furthermore, absent the WeakReferences TC39 proposal, shims must be manually freed or memory leaks could ensue.

Ultimately, given that this whole proposal is just a stop-gap until WASM interface-types lands and there wasn't a huge amount of interest, I wasn't convinced it was worth pursuing?  I did try breaking it down in "building blocks" as @alexcrichton suggested, but I struggle to view this as much more than the one "big" change to shim construction (with everything else being housekeeping required to make that work).

PR #12, opened 2020-02-14
Better error handling

Pauan:
[Rendered](https://github.com/Pauan/rfcs/blob/better-error-handling/text/009-error-handling.md)

### DIFF
--- a/text/009-error-handling.md
+++ b/text/009-error-handling.md
@@ -32,7 +32,7 @@ call must use `try`/`catch` to convert the error into `Result`. And this also
 bloats up the generated JS glue file, because the `try`/`catch` cannot really
 be deduplicated.
 
-Lastly, the current system means we must always use JS glue code,
+The current system also means we must always use JS glue code,
 even after interface types are implemented in the browsers.
 
 And all of the above costs are unavoidable: even if you *know* that it will never
@@ -44,7 +44,7 @@ so even though the costs may be small for a single function, it adds up with
 hundreds or thousands of functions.
 
 As a more meta note, it is idiomatic in Rust to return custom error types,
-such as `Result<T, MyCustomError>`. However the current system heavily
+such as `Result<T, MyCustomError>`. However, the current system heavily
 encourages users to *not* do that and instead just return `Result<T, JsValue>`.
 
 The reason is because if you return `Result<T, JsValue>` then you can use the
@@ -158,7 +158,7 @@ However, it fixes all of the issues described in the motivation:
 
 * With the current system, if the author did not use the `catch`
    attribute then you cannot catch the error. But with `try_catch` you
-   can! Thus this is more flexible.
+   can! Thus it is more flexible.
 
 
 # Drawbacks
@@ -186,7 +186,7 @@ However, it fixes all of the issues described in the motivation:
 
    Also, the current system of always returning `JsValue` is really terrible
    for error handling. It is far better to create custom errors, which is
-   something that gloo can do.
+   discouraged by the current system.
 
 
 # Rationale and Alternatives
--- a/text/009-error-handling.md
+++ b/text/009-error-handling.md
@@ -1,0 +1,219 @@
+- Start Date: 2020-02-14
+- RFC PR: (leave this empty)
+- Tracking Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Dramatically improve the ergonomics of error handling by adding a new
+`try_catch` function and changing JS functions to no longer return `Result`.
+
+
+# Motivation
+[motivation]: #motivation
+
+Right now almost all APIs in the `wasm-bindgen` ecosystem return
+`Result<T, JsValue>`. This is true even in the cases where the API cannot error.
+
+This causes a large amount of friction when using the APIs, because it is
+necessary to either propagate the error or use `.unwrap_throw()`.
+
+In addition, Rust has the policy that programmer errors should be handled via
+`panic`, but runtime failures should be handled by `Result`. This distinction has
+also been noticed in the JavaScript community, as shown in this 2014 article:
+
+https://www.joyent.com/node-js/production/design/errors
+
+But the current system with `wasm-bindgen` conflates those two together, with
+programmer errors being handled via `Result`.
+
+The current system also has a small performance cost, since every function/method
+call must use `try`/`catch` to convert the error into `Result`. And this also
+bloats up the generated JS glue file, because the `try`/`catch` cannot really
+be deduplicated.
+
+Lastly, the current system means we must always use JS glue code,
+even after interface types are implemented in the browsers.
+
+And all of the above costs are unavoidable: even if you *know* that it will never
+error, you must still pay the cost in JS glue, file size, and runtime performance.
+
+This means `wasm-bindgen` is no longer truly zero-cost. And unfortunately
+this problem is pervasive throughout the entire `wasm-bindgen` ecosystem,
+so even though the costs may be small for a single function, it adds up with
+hundreds or thousands of functions.
+
+As a more meta note, it is idiomatic in Rust to return custom error types,
+such as `Result<T, MyCustomError>`. However the current system heavily
+encourages users to *not* do that and instead just return `Result<T, JsValue>`.
+
+The reason is because if you return `Result<T, JsValue>` then you can use the
+`?` syntax, which is far more ergonomic. So by changing the error handling
+system we can encourage users to create meaningful custom errors, rather
+than `Result<T, JsValue>`.
+
+
+# Stakeholders
+[stakeholders]: #stakeholders
+
+This affects basically the entire `wasm-bindgen` ecosystem, since it is a
+breaking change for the `js-sys` and `web-sys` crates.
+
+I'm not sure what approach is appropriate for gaining feedback.
+
+
+# Detailed Explanation
+[detailed-explanation]: #detailed-explanation
+
+The current system looks like this:
+
+```rust
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(catch)]
+    fn foo() -> Result<u32, JsValue>;
+}
+
+fn bar() {
+    foo().unwrap_throw();
+}
+```
+
+This generates the following JS glue code:
+
+```js
+imports.wbg.__wbg_foo_12a2d9a2dfcf125f = function() {
+    try {
+        var ret = foo();
+        return ret;
+    } catch (e) {
+        handleError(e);
+    }
+};
+```
+
+What I propose is to remove the `catch` attribute and no longer
+have `wasm_bindgen` functions return `Result`.
+
+Instead, wasm-bindgen would define a new `try_catch` function, with
+this signature:
+
+```rust
+fn try_catch<A, F>(f: F) -> Result<A, JsValue>
+    where F: FnOnce() -> A + UnwindSafe
+```
+
+This function would call a new primitive, which basically does this:
+
+```js
+function try_catch(f) {
+    try {
+        return f();
+    } catch (e) {
+        handleError(e);
+    }
+}
+```
+
+(It will actually be implemented differently, because it has to call a
+Rust closure, not a JS function, but that's implementation details.)
+
+Now it is possible to catch JS errors and convert them into `Result`:
+
+```rust
+#[wasm_bindgen]
+extern "C" {
+    fn foo() -> u32;
+}
+
+fn bar() {
+    try_catch(|| foo()).unwrap_throw();
+}
+```
+
+This has the same behavior as the earlier code which uses the `catch`
+attribute.
+
+However, it fixes all of the issues described in the motivation:
+
+* Users don't need to use `try_catch` in the situations where they know
+   that the call won't error.
+
+* Users don't need to use `try_catch` in the situations where the error
+  is a programmer error and not a runtime failure.
+
+* Users can put multiple calls within the `try_catch` closure, thus
+   amortizing the runtime costs.
+
+* The `try`/`catch` functionality only needs to be defined once,
+   rather than repeated in every `catch` function.
+
+* The JS glue code is significantly reduced, and with interface types
+   it can be completely eliminated.
+
+* With the current system, if the author did not use the `catch`
+   attribute then you cannot catch the error. But with `try_catch` you
+   can! Thus this is more flexible.
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+* This is a major breaking change to the entire ecosystem. It will obviously
+   need major version bumps in `wasm-bindgen`, `js-sys`, `web-sys`, and any
+   crates which transitively use them.
+
+   Therefore, if this is accepted we should make the change ASAP, so we can
+   minimize the pain of transitioning. It's better to make a large change like
+   this now, rather than later when there are more people using Wasm.
+
+* Another drawback is that this can encourage users to not handle errors at
+   all, because the APIs no longer return `Result`.
+
+   However, in practice users already don't handle errors (they just use
+   `.unwrap_throw()`), and it is difficult to properly handle errors in JS
+   anyways, because you need to do string parsing to figure out what the
+   error is.
+
+   So if we want to encourage users to properly handle errors, I think
+   there are far better ways to do that, such as providing excellent
+   APIs for parsing and handling errors.
+
+   Also, the current system of always returning `JsValue` is really terrible
+   for error handling. It is far better to create custom errors, which is
+   something that gloo can do.
+
+
+# Rationale and Alternatives
+[alternatives]: #rationale-and-alternatives
+
+Obviously we can choose to not do this, but that means accepting all of
+the problems in the Motivation forever.
+
+I'm unaware of any other alternatives.
+
+
+# Future Extensions
+[future]: #future-extensions
+
+N/A
+
+
+# Unresolved Questions
+[unresolved]: #unresolved-questions
+
+The implementation will be a little tricky, it basically involves boxing
+the `FnOnce` closure, sending the pointer to JS, then having JS call back
+into Wasm with that pointer.
+
+It's also an open question how things like Rust panics will be handled.
+I added `UnwindSafe` to the signature for `try_catch`, which is *probably*
+good enough, but I haven't thought particularly hard about this.
+
+And right now there isn't good interop between Wasm traps and JS exceptions,
+and the Wasm exceptions standard is still up in the air. So we don't really know
+how all of that will play out.
+
+However, I imagine that eventually Wasm will get first-class support for
+unwinding and catching JS exceptions, and the `try_catch` function will be
+able to seamlessly use that in the future.
--- a/text/009-error-handling.md
+++ b/text/009-error-handling.md
@@ -12,7 +12,7 @@ Dramatically improve the ergonomics of error handling by adding a new
 # Motivation
 [motivation]: #motivation
 
-Right now almost all APIs in the `wasm-bindgen` ecosystem return
+Right now almost all APIs in the wasm-bindgen ecosystem return
 `Result<T, JsValue>`. This is true even in the cases where the API cannot error.
 
 This causes a large amount of friction when using the APIs, because it is
@@ -24,7 +24,7 @@ also been noticed in the JavaScript community, as shown in this 2014 article:
 
 https://www.joyent.com/node-js/production/design/errors
 
-But the current system with `wasm-bindgen` conflates those two together, with
+But the current system with wasm-bindgen conflates those two together, with
 programmer errors being handled via `Result`.
 
 The current system also has a small performance cost, since every function/method
@@ -38,8 +38,8 @@ even after interface types are implemented in the browsers.
 And all of the above costs are unavoidable: even if you *know* that it will never
 error, you must still pay the cost in JS glue, file size, and runtime performance.
 
-This means `wasm-bindgen` is no longer truly zero-cost. And unfortunately
-this problem is pervasive throughout the entire `wasm-bindgen` ecosystem,
+This means wasm-bindgen is not truly zero-cost. And unfortunately
+this problem is pervasive throughout the entire wasm-bindgen ecosystem,
 so even though the costs may be small for a single function, it adds up with
 hundreds or thousands of functions.
 
@@ -52,11 +52,16 @@ The reason is because if you return `Result<T, JsValue>` then you can use the
 system we can encourage users to create meaningful custom errors, rather
 than `Result<T, JsValue>`.
 
+Also, in the past we had made decisions to adopt a more JS-centric API
+(such as using `Deref` rather than traits). This proposal is in line with
+that, because it is idiomatic in JS to only use `try`/`catch` when needed,
+rather than on every function call.
+
 
 # Stakeholders
 [stakeholders]: #stakeholders
 
-This affects basically the entire `wasm-bindgen` ecosystem, since it is a
+This affects basically the entire wasm-bindgen ecosystem, since it is a
 breaking change for the `js-sys` and `web-sys` crates.
 
 I'm not sure what approach is appropriate for gaining feedback.


Pauan:
P.S. I was worried that this proposal wouldn't work, due to the weird interaction between Wasm and JS exceptions.

But I ran some tests, and it seems to work fine: a JS exception triggered by a Rust closure is still caught correctly, even though it passes through the Wasm<->JS bridge *twice*. And as far as I can tell it doesn't have any problems with invalid state.

Even panicking seems to work (though it catches the `RuntimeError: unreachable` rather than the actual panic, which is expected).

However, any Rust state created inside of the closure won't have its destructors run if an exception occurs:

```rust
try_catch(|| {
    // It won't call the Drop impl for x
    let x = some_type_with_drop();
    trigger_js_exception();
});
```

That's fine, `try_catch` is intended for *JS* APIs, so you shouldn't be doing that anyways. But this does need to be documented.

Pauan:
P.P.S. This is the implementation I've been using for my experiments:

```rust
use std::panic::UnwindSafe;

#[wasm_bindgen(inline_js = "
    export function try_catch_(f) {
        try {
            f();

        } catch (e) {
            return e;
        }
    }
")]
extern "C" {
    fn try_catch_(f: &mut dyn FnMut()) -> JsValue;
}

fn try_catch<A, F>(f: F) -> Result<A, JsValue>
    where F: FnOnce() -> A + UnwindSafe {

    // We have to use Box so that way we can manually free it if an error occurs.
    let pointer = Box::into_raw(Box::new(f));

    let mut value = None;

    let error = try_catch_(&mut || {
        let f = unsafe { Box::from_raw(pointer) };
        value = Some(f());
    });

    // It didn't error, and therefore the Box has already been freed.
    if let Some(value) = value {
        Ok(value)

    } else {
        // This is safe because if an error occured then the Box was not dropped.
        let _ = unsafe { Box::from_raw(pointer) };
        Err(error)
    }
}
```

It's slower than it could be, but it works correctly, and it doesn't require any changes to wasm-bindgen.

This is the code I've been using to test it:

```rust
#[wasm_bindgen(start)]
pub fn main_js() {
    console_error_panic_hook::set_once();

    struct Dropper(&'static str);

    impl Drop for Dropper {
        fn drop(&mut self) {
            console::log_1(&JsValue::from(format!("Dropping {}", self.0)));
        }
    }

    let _top = Dropper("top");

    for _ in 0..1_000_000 {
        let _ = try_catch(move || {
            let _y: [u32; 10_000] = [0; 10_000];
            let _closure = Dropper("closure");
            wasm_bindgen::throw_str("Error!!");
        });
    }

    console::log_1(&JsValue::from("Everything is fine!"));
}
```

This verifies that it correctly frees up the stack space, but doesn't run destructors for types created inside of the closure.

It also verifies that the state of both JS and Rust is at least somewhat sane.

Pauan:
As an anecdote, in one of my projects the error handling alone is currently taking up 1,205 bytes in the ***minified*** JS glue code, and another project is taking up 2,410 bytes.

That would be completely eliminated by this proposal, making my projects *kilobytes* smaller.

alexcrichton:
Looks like a reasonable proposal, thanks for writing this up! I'm generally fine with PRs/issues for discussion FWIW, I don't think the RFC process has worked out that well for us (nor has it really been all that necessary).

My main concern with this would be the stack pointer and https://github.com/rustwasm/wasm-bindgen/issues/1963. I see you've written out an example that seems to work, but given what I know about our codegen I don't think it's correct? AFAIK there's nothing to restore the stack pointer, so it's pointing off into the middle of nowhere. My guess it that it somewhat works because each function keeps a `local` with its own value of the stack pointer, so it doesn't keep reloading the stack pointer for local calculations. If you call the function in a loop though I think it will have the same bug as before (leaking stack space)

Pauan:
> If you call the function in a loop though I think it will have the same bug as before (leaking stack space)

I'll need to do more tests, but even if I put things into a separate non-inlined function which creates a lot of stack space, it still works fine, without any errors:

```rust
#[inline(never)]
fn my_fn() -> Result<(), JsValue> {
    let _y: [u32; 10_000] = [0; 10_000];

    try_catch(move || {
        let _y: [u32; 10_000] = [0; 10_000];
        wasm_bindgen::throw_str("Error!!");
    })
}


#[wasm_bindgen(start)]
pub fn main_js() {
    console_error_panic_hook::set_once();

    for _ in 0..1_000_000 {
        let _ = my_fn();
    }

    console::log_1(&JsValue::from("Everything is fine!"));
}

```

I think the situation is similar but different from https://github.com/rustwasm/wasm-bindgen/issues/1963

Though if you have any ideas for good stress tests, I'd like to run them.

alexcrichton:
I'd recommend looking at the generated `*.wat` because it's easy for those sorts of allocations to get optimized away. You'll want to double-check that the closure here actually generates stack space which won't get reclaimed when the exception is caught.

Pauan:
So I did everything I could possibly think of to guarantee that it would be included:

```rust
#[wasm_bindgen(inline_js = "
    export function error() {
        throw new Error(\"Hello world!!!\");
    }
")]
extern "C" {
    fn error() -> JsValue;
}

#[inline(never)]
fn data() -> [f64; 10_000] {
    [js_sys::Math::random(); 10_000]
}

#[inline(never)]
fn check_data(a: &[f64; 10_000]) {
    assert_eq!(a.len(), 10_000);

    let mut counter = 0;

    let first = a[0];

    for &elem in a.iter() {
        counter += 1;
        assert!(elem >= 0.0);
        assert!(elem <= 1.0);
        assert_eq!(first, elem);
    }

    assert_eq!(counter, 10_000);
}

#[inline(never)]
fn my_fn() -> Result<(), JsValue> {
    let y = std::hint::black_box(data());
    check_data(&y);

    try_catch(move || {
        let y = std::hint::black_box(data());
        check_data(&y);
        error();
    })
}

#[wasm_bindgen(start)]
pub fn main_js() {
    console_error_panic_hook::set_once();

    for _ in 0..1_000_000 {
        let _ = my_fn();
    }

    console::log_1(&JsValue::from("Everything is fine!"));
}
```

* It uses `Math.random()` so it isn't a constant, so Rust can't optimize that.

* It uses `black_box` to try and remove as many optimizations as possible.

* It puts in runtime checks which verify that the array exists and has the right elements (these runtime checks can't be eliminated by Rust because they rely on the actual value of the array, which is random).

* It uses a custom `error` function, so that way Rust can't do any optimizations with `!` (which is returned by `throw_str`).

But it *still* works fine.

From examining the `.wat`, it is actually creating the array. And looking at the `.wat` for the closure, it is indeed leaking stack space after the closure call, *but* it then restores the correct stack pointer at the end of the `my_fn` call.

Since closure calls must always be done indirect (they cannot be inlined), I *believe* this guarantees that the stack pointer will be correctly restored under all circumstances. But it's hard to prove that in a satisfactory way.

Pauan:
Okay, I was *finally* able to trigger a stack leak:

```rust
#[wasm_bindgen(inline_js = "
    export function error() {
        throw new Error(\"Hello world!!!\");
    }
")]
extern "C" {
    fn error() -> JsValue;
}

#[inline(never)]
fn data() -> [f64; 10_000] {
    [js_sys::Math::random(); 10_000]
}

#[inline(never)]
fn check_data(a: &[f64; 10_000]) {
    assert_eq!(a.len(), 10_000);

    let mut counter = 0;

    let first = a[0];

    for &elem in a.iter() {
        counter += 1;
        assert!(elem >= 0.0);
        assert!(elem <= 1.0);
        assert_eq!(first, elem);
    }

    assert_eq!(counter, 10_000);
}

#[wasm_bindgen(start)]
pub fn main_js() {
    console_error_panic_hook::set_once();

    let y = std::hint::black_box(data());
    check_data(&y);

    for _ in 0..1_000_000 {
        assert!(try_catch(move || {
            let y = std::hint::black_box(data());
            check_data(&y);
            error();
        }).is_err());
    }

    console::log_1(&JsValue::from("Everything is fine!"));
}
```

This is because it no longer uses the `my_fn` function. Since the stack pointer is restored at the end of every function, and there aren't any more Rust function calls within the loop, it never restores the stack pointer.

And of course the stack pointer *is* restored correctly at the end of `main_js` (or whatever function contains the loop).

This is a very contrived situation which seems hard to trigger and is ***extremely*** unlikely to happen in practice, but we should still try to find a solution for it.

Pauan:
I was able to fix it by simply not inlining `try_catch`:

```rust
#[inline(never)]
fn try_catch<A, F>(f: F) -> Result<A, JsValue>
    where F: FnOnce() -> A + UnwindSafe {
    // ...
}
```

This works because it will restore the stack pointer at the end of the `try_catch` function.

This is relying on the implementation detail that functions always restore a local copy of the stack pointer, but I don't see a better solution, at least until Wasm exceptions is available.

Pauan:
As an alternative, we can wrap the `try_catch_` in a non-inline identity function, like this:

```rust
/// This is used to restore the correct stack pointer after a function call
#[inline(never)]
fn id<A, F>(f: F) -> A where F: FnOnce() -> A { f() }

fn try_catch<A, F>(f: F) -> Result<A, JsValue>
    where F: FnOnce() -> A + UnwindSafe {

    // We have to use Box so that way we can manually free it if an error occurs.
    let pointer = Box::into_raw(Box::new(f));

    let mut value = None;

    let error = id(|| {
        try_catch_(&mut || {
            // This is safe because this closure will only be called once
            let f = unsafe { Box::from_raw(pointer) };
            value = Some(f());
        })
    });

    // It didn't error, and therefore the Box has already been freed.
    if let Some(value) = value {
        Ok(value)

    } else {
        // This is safe because if an error occured then the Box was not dropped.
        let _ = unsafe { Box::from_raw(pointer) };
        Err(error)
    }
}
```

This allows the `try_catch` function to still be inlined, but it correctly restores the stack pointer after the `try_catch_` call.

So the above implementation doesn't leak stack, and it works correctly under all circumstances, except that it doesn't run destructors on values created *inside* the closure (but it still runs destructors on values *captured* by the closure).

The `id` function generates code like this:

```wat
  (func $f97 (type $t0) (param $p0 i32) (param $p1 i32) (result i32)
    (local $l0 i32)
    (set_global $g0
      (tee_local $l0
        (i32.sub
          (get_global $g0)
          (i32.const 16))))
    (i32.store offset=12
      (get_local $l0)
      (get_local $p1))
    (i32.store offset=8
      (get_local $l0)
      (get_local $p0))
    (call $wbg.__wbg_trycatch_8e76280895c219ba
      (i32.add
        (get_local $l0)
        (i32.const 8))
      (i32.const 1048576))
    (set_global $g0
      (i32.add
        (get_local $l0)
        (i32.const 16))))
```

This seems close to optimal to me: it just calls the `try_catch_` function and then restores the stack pointer afterwards. So it's doing the minimum work necessary.

We may be able to use this technique to fix https://github.com/rustwasm/wasm-bindgen/issues/1963 as well.

yoshuawuyts:
### Async

Something I'm missing in this proposal is the interaction with async code. When a new Promise is spawned in JS, the try-catch block will no longer catch errors bubbling out of it. Only when used with async/await in JS will this interaction work as intended; but that's currently not the abstraction `spawn_local` provides. Given the proposed `try_catch` function takes a synchronous closure in Rust I'm unsure how that interaction is intended to work.

### Abstractions

I'd like to note that in my experience `.unwrap_throw()` has proven to be enough to provide efficient abstractions. As we're building out abstractions that provide higher-level APIs these details are not exposed to end-users either way, so I'm not too concerned about details at this layer.


### Compression

Also regarding size on disk: if the JS bundle size stems from repeated use of `try..catch`, then it's probably also worth measuring size over the wire _after_ compression. The impact of this code is likely significantly smaller then when only measuring minified JS code.

Pauan:
> @yoshuawuyts Something I'm missing in this proposal is the interaction with async code.

That seems orthogonal and unrelated to me.

Promises follow the same semantics they do in JS: they completely swallow all errors and convert them into a Promise rejection, which is why `JsFuture::from` always returns a `Result`. That doesn't change with this proposal (and it can't really change, given the nature of JS).

So there aren't any changes needed for async JS code.

> but that's currently not the abstraction spawn_local provides.

I don't really understand your point: `spawn_local` is an abstraction for *Rust Futures*, not JS Promises. So it has the normal semantics that Rust Futures have (which do not catch panics/errors).

So maybe you meant the interaction that `future_to_promise` should have?

> Given the proposed try_catch function takes a synchronous closure in Rust I'm unsure how that interaction is intended to work.

Promises already catch errors and return `Result`, that doesn't change.

As for synchronous errors when calling a function that returns a Promise, in practice those are almost always programmer errors.

But in the rare case where they aren't, you can use `try_catch` as usual, it behaves the same as `try`/`catch` in JS, so you would use it exactly the same way.

> Also regarding size on disk: if the JS bundle size stems from repeated use of try..catch, then it's probably also worth measuring size over the wire after compression.

Though I'm fine with measuring the impact of compression, my projects cannot be compressed (they are Chrome/Firefox Extensions). Not all uses of JS are over the web.

And although compression may help with the download time, the bytes still have to be uncompressed (taking up CPU and RAM), and the final uncompressed bytes have to be loaded and executed. In addition, the uncompressed bytes will be sitting in browser/server caches, which have a fixed size, so larger files will cause cache eviction. So the uncompressed size does matter.

I don't think it's good to rely on compression to compensate for bloated code, especially when the problem can be fixed completely. Even more so since the code is auto-generated and cannot be avoided by the user.

alexcrichton:
Ah ok @Pauan what you've found aligns with my understanding of how things should work, and I can see how an `#[inline(never)]` tiny shim would "fix" the issue.

I'm a bit wary to rely on such fragile codegen details like this, but it's also probably not the end of the world. I'd want to probably dig in a bit closer with the intricacies of what's happening here at some point before landing this, but I have a hunch that it'd work well enough today to get the job done.

I do think though that there's one other thing worth bringing up which I've forgotten about. Technically it is memory unsafe to unwind stack frames and not execute destructors on the stack frames. For example scoped threads in Rust are only safe because if the stack is unwound then destructors are run. In general it's acceptable to not run destructors in Rust, but if a value is owned by the stack and the stack is unwound it's not memory safe to skip running those destructors. (that's a case where Rust guarantees that destructors are run today)

This is a pretty uncomfortable question for wasm, and one we haven't mostly just ignored up to now. Our general mantra has been "well just catch JS exceptions entering wasm and propagate the `Result`". With a function like `try_catch`, however, that makes the error-handling pattern much more prominent and makes it much easier (I think?) to leak resources in Rust, possibly to the detriment of memory safety. 

Even without the memory safety question though this makes is very easy to leak resources in Rust which generally wants to be avoided as well. Unfortunately though there's no real way to square this other than "add a full gc" or "wait for the exceptions proposal". I'm curious on your thoughts on this though? Especially wrt the ease that resources can be leaked now.

Pauan:
> I'm a bit wary to rely on such fragile codegen details like this, but it's also probably not the end of the world.

I fully agree, but it's a temporary measure until Wasm exceptions are available, and we can have unit tests to verify that it's not leaking stack space, so we can catch any regressions.

> Technically it is memory unsafe to unwind stack frames and not execute destructors on the stack frames.

Doesn't that mean that wasm-bindgen is already memory unsafe? Since any JS API which isn't marked as `catch` can throw an exception which breaks the stack pointer and doesn't run stack destructors.

> With a function like try_catch, however, that makes the error-handling pattern much more prominent and makes it much easier (I think?) to leak resources in Rust, possibly to the detriment of memory safety.

As far as I know, that's only true if you create things inside of the closure, so `try_catch(|| foo(bar))` should be safe, right?

> Even without the memory safety question though this makes is very easy to leak resources in Rust which generally wants to be avoided as well. Unfortunately though there's no real way to square this other than "add a full gc" or "wait for the exceptions proposal". I'm curious on your thoughts on this though?

My solution is to just make it ***very*** clear in the docs that this can leak things which are created inside of the closure. So it's best to make the `try_catch` closure as small as possible.

If we want to, we can make `try_catch` `unsafe`, which may be justified if it can cause memory unsafety.

alexcrichton:
Yes wasm-bindgen is already technically memory unsafe due to this issue. It's not really a problem in practice since if you want to catch something you "catch and throw" at the boundary most of the time rather than in the middle of some code. The `throw_str` and `throw_val` functions in wasm-bindgen should be (and are not AFAIK) clearly documented that they leak and should be highly discoraged from use.

The general idea as well is that if an exception is thrown and you didn't expect it then you probably won't reenter the wasm, but we also do not place any guards anywhere preventing you from reentering the wasm (but we arguably should).

> As far as I know, that's only true if you create things inside of the closure, so try_catch(|| foo(bar)) should be safe, right?

This is both sort of right but also sort of wrong. For example if `bar` is an owned value then it will leak if `foo` throws an exception and didn't drop `bar` first. Furthermore this could easily be accompanied with a lot of stack frames between the `try_catch` and the actual exit into JS that then caught the exception. I think given how ergonomic `try_catch` will be it's going to be very easy to get buried deep in a library so that callers don't realize they're leaking destructors on the stack.

I think it's a given the documentation will be clear about this, but I think that I would want to mark it as `unsafe` in the meantime as well. That will hopefully strongly discourage users from using it deep within libraries, and it would also help get folks reading the documentation to read about the caveats.

---

As one sort of minor point, we likely don't want this as a blanket `try`/`catch` but rather we'll likely want to do some filtering. In discussions at the CG it was pointed out that even the exception handling proposal won't catch wasm traps such as out of bounds reads/writes or stack overflow. We'll want to be sure to not catch these exceptions as well and, if caught, reraise them.

Pauan:
> It's not really a problem in practice since if you want to catch something you "catch and throw" at the boundary most of the time rather than in the middle of some code.

To be clear, I'm not talking about catching, I'm saying that *any* JS API can throw, and that can screw things up. For example...

```rust
#[wasm_bindgen]
extern "C" {
    fn some_function_that_throws();
}

// The JS setTimeout function "swallows" errors without stopping execution.
// This happens with *all* async functions, including Promises, event
// listeners, etc.
set_timeout(|| {
    let x = allocate_something();
    some_function_that_throws();
}, 0);
```

In this case it won't run the destructor for `x`. So *every* JS API can potentially cause this issue, even without catching.

So if we wanted to be super correct we would make all JS APIs `unsafe`, but that seems like going too far.

> The general idea as well is that if an exception is thrown and you didn't expect it then you probably won't reenter the wasm, but we also do not place any guards anywhere preventing you from reentering the wasm (but we arguably should).

I agree, but that seems rather difficult to enforce without significant performance penalties.

Given that, I think it's reasonable for us to say, "if any uncaught error happens, all bets are off, so make sure to fix all errors".

I think that isn't too different from how Rust deals with C FFI: unwinding doesn't work between Rust and C++, so you have to be extra careful when dealing with C++ errors. Basically, Rust just kinda gives up and says "here be dragons", since it can't really fix the issue.

> For example if bar is an owned value then it will leak if foo throws an exception and didn't drop bar first.

Hmmm, from my testing that doesn't seem to be the case:

```rust
let x = Dropper("owned");

fn foo(x: Dropper) {
    error();
}

assert!(try_catch(move || {
    foo(x);
}).is_err());
```

I had carefully implemented it so that it will drop all *captured* values, so it's only values created inside the closure which aren't dropped.

Though, with some further testing, it seems to be possible to trigger a double drop, which is quite bad. So maybe it's better for it to *not* drop captured values, and leave it up to the user to do that.

> I think it's a given the documentation will be clear about this, but I think that I would want to mark it as unsafe in the meantime as well. That will hopefully strongly discourage users from using it deep within libraries, and it would also help get folks reading the documentation to read about the caveats.

I agree with making it `unsafe`, but I imagine that aside from converting JS errors into nice custom Rust errors, there just isn't that much use for `try_catch`, so hopefully it won't be a problem in practice.

> As one sort of minor point, we likely don't want this as a blanket try/catch but rather we'll likely want to do some filtering.

That seems brittle, since it will be relying on string parsing, which might vary from browser to browser.

From some light testing, it seems that out of bounds exceptions are *already* uncatchable by JS, so the behavior might already be correct, so we don't need to do any filtering.

alexcrichton:
Honestly I'm not sure how to best thread this needle, we in general don't really have a lot of great options if you want to catch JS exceptions in Rust right now, and our only "real" recourse is the exceptions proposal which is still aways away. There's various points we can debate sorta but it really just all boils down to that JS exceptions don't run Rust destructors, and how much can we get away with despite that. 

I'm personally for being pretty conservative, basically adding this as an `unsafe` API. It's not *impossible* to use incorrectly, it's just hard :)

ctaggart:
WebAssembly Exception handling proposal is what I think Alex is referring to:
https://github.com/WebAssembly/exception-handling/blob/master/proposals/Exceptions.md

Pauan:
@alexcrichton So I've given some thought to this, I think I have a tentative solution, but I'm a bit squeamish due to the performance costs.

Since [*any* non-`catch` JS function is currently unsafe](https://github.com/rustwasm/rfcs/pull/12#issuecomment-589457200) (in the exact same way as `try_catch`), we should try to solve that as well.

So here is my proposal:

1. JS functions no longer return `Result`.

2. wasm-bindgen will catch *all* exceptions for *all* JS functions (even the non-`catch` ones).

3. We create a global "exception handler" for wasm-bindgen, similar to `std::panic::set_hook` (except for JS exceptions).

4. By default this exception handler will trigger a Rust panic, which means that the stack will be unwound correctly (!).

5. But the `try_catch` function can override the exception handler, allowing it to catch the error and return `Result`.

6. After Wasm exceptions are available, we can use that instead of the exception handler, as an optimization.

This is essentially a polyfill for Wasm exceptions, which has the following performance costs:

* Functions which are currently non-`catch` will become `catch`.

* It has to check the global exception handler on every JS call.

Those performance costs will go away after Wasm exceptions arrive, but in the meantime we have correct behavior (including stack unwinding).

Most web-sys APIs are already `catch`, so they won't suffer too much, but the non-web-sys APIs will be hit harder, so we'll need to run some benchmarks.

Pauan:
By the way, since `catch` would be the default, we should add a new `unsafe_never_throws` attribute, which means the function is guaranteed to never throw a JS exception.

This means wasm-bindgen doesn't need to catch the exception, and it doesn't need to run the exception handler.

We can then selectively add that attribute to the APIs which we know won't throw, and thus avoid the performance cost. Adding the attribute to a function is a non-breaking change, since it doesn't affect the type, it's purely an optimization.

Pauan:
I've created a [working prototype](https://github.com/rustwasm/wasm-bindgen/pull/2093) of `try_catch` implemented using the global exception handler approach. It was a *lot* easier than trying to implement `try_catch` natively.

It actually works great: stack variables are cleaned up, and the closure itself is also cleaned up, so there is no leak and no undefined behavior!

There is *one* issue however: when a JS function throws an exception inside of the `try_catch`, it will not *stop* execution, which means...

```rust
try_catch(|| {
    let x = function_that_throws();
    some_other_function(x)
})
```

If `function_that_throws()` throws an exception, then `x` will be `JsValue::UNDEFINED` and it will call `some_other_function`. This problem only happens *inside* `try_catch`. This behavior is annoying, but it isn't unsound, so it's just a minor footgun.

This problem can be fixed by simply calling only one function inside of `try_catch`:

```rust
let x = try_catch(|| function_that_throws())?;
some_other_function(x)
```

After Wasm exceptions are available we can fully fix this properly, but until then we'll just have to accept it. But all of the unsound behavior and memory leaks are fixed.

Also, I implemented the global exception handler in an efficient way: on the happy path it has no extra cost, and the cost on the error path is *fairly* small. So in practice I expect this strategy to have similar performance to the current status quo.

I think this is a viable solution, but we still have to do more benchmarks.

Pauan:
I've done some benchmarks, the performance cost is *extremely* tiny, it's so small that it's within the level of noise:

```
Without catch:  0.000033955815032611126 ms
With catch:     0.000038553162430695450 ms
```

The difference between them is ~5 ***nano***seconds. I still need to do some more benchmarks on a full app, but thus far it seems that the performance cost is basically nonexistent. So it's really starting to seem like this implementation is viable.

alexcrichton:
An interesting idea! 

I'm a bit hazy on some of the parts though. You say the stack will be unwound correctly and the diff mentions that [even if `f` panics it resets the handler](https://github.com/rustwasm/wasm-bindgen/pull/2093/files#diff-b4aea3e418ccdb71239b96952d9cddb6R1073), but how is that all possible? I think that's only true if you're in a `try_catch`, right?

This also looks like it may be sort of iffy for other reasons due to the return of a "zero value" when an exception otherwise happens. When an exception is stored in the JS glue I don't think any `return` happens, right? That I believe means that if wasm code continues after the exception is thrown then it'll read stale data. For example if you do this:

```rust
#[wasm_bindgen]
extern "C" {
    fn foo() -> SomeRustStructure;
}

fn foo() {
    try_catch(|| {
        let x: SomeRustStructure = foo();
        // `x` has a null pointer internally
    })
}
```

I think that can cause runtime corruption? This may be worse for functions that return things like `&str`. Basically what I'm getting at is that I believe with `try_catch` it doesn't actually interrupt control flow like you'd expect on an exception, right? It looks like it allows continuing execution in wasm as usual, despite the exception happening.

Pauan:
@alexcrichton The way that it works is:

1. [*All* JS calls will now catch](https://github.com/rustwasm/wasm-bindgen/blob/8d09c214d2e22f58290ea41292218e051c850a70/crates/macro-support/src/parser.rs#L507) (even if they're not marked as `catch`). This seems to have a [negligible performance cost](https://github.com/rustwasm/rfcs/pull/12#issuecomment-617488215).

2. When calling a JS API, rather than returning `Result`, it instead [checks](https://github.com/rustwasm/wasm-bindgen/blob/8d09c214d2e22f58290ea41292218e051c850a70/crates/backend/src/codegen.rs#L990) the [exception handler](https://github.com/rustwasm/wasm-bindgen/blob/8d09c214d2e22f58290ea41292218e051c850a70/src/lib.rs#L1041).

3. It just does a [quick check](https://github.com/rustwasm/wasm-bindgen/blob/8d09c214d2e22f58290ea41292218e051c850a70/src/lib.rs#L1135) to see if there is an exception (which should be very fast), so there is essentially no cost if there isn't an error. But if there is an error, it then [calls the exception handler](https://github.com/rustwasm/wasm-bindgen/blob/8d09c214d2e22f58290ea41292218e051c850a70/src/lib.rs#L1034).

4. The default exception handler [does a panic](https://github.com/rustwasm/wasm-bindgen/blob/8d09c214d2e22f58290ea41292218e051c850a70/src/lib.rs#L1030), which will unwind the stack (well, right now unwinding isn't supported on Wasm, but when it *is* supported then it will unwind the stack).

5. The `try_catch` function [overwrites the exception handler](https://github.com/rustwasm/wasm-bindgen/blob/8d09c214d2e22f58290ea41292218e051c850a70/src/lib.rs#L1071) with an implementation which [doesn't panic but instead just stores the exception](https://github.com/rustwasm/wasm-bindgen/blob/8d09c214d2e22f58290ea41292218e051c850a70/src/lib.rs#L1068).

   After the `f()` function runs it will [reset the exception handler back to the original](https://github.com/rustwasm/wasm-bindgen/blob/8d09c214d2e22f58290ea41292218e051c850a70/src/lib.rs#L1058). This uses `Drop` so that way it will always reset the exception handler even if `f()` panics.

   This is basically the Rust version of `try`/`finally`: we change the exception handler, run `f()`, then reset the exception handler regardless of whether `f()` panics or not.

   Lastly `try_catch` will [check the stored exception](https://github.com/rustwasm/wasm-bindgen/blob/8d09c214d2e22f58290ea41292218e051c850a70/src/lib.rs#L1081) to see if an exception happened.

So to summarize: all JS APIs will now catch errors (even those which aren't marked with `catch`). If an error happens it calls the thread-local exception handler (which by default will `panic!`).

The `try_catch` function temporarily overwrites the exception handler so that way it can catch the exception and store it (rather than panicking).

So regardless of whether you use `try_catch` or not, it will be catching and checking all JS APIs, and calling the exception handler if an error happens. The only difference with `try_catch` is that the behavior of the exception handler is different.

Pauan:
> When an exception is stored in the JS glue I don't think any return happens, right?

When a JS function throws an exception, it will always return `undefined`:

```js
imports.wbg.__wbg_foo_f181bd863f0ccb9d = function() {
    try {
        var ret = foo();
        return ret;
    } catch (e) {
        wasm.__wbindgen_exn_store(addHeapObject(e));
    }
};
```

That's because the `catch` swallows the exception.

> I think that can cause runtime corruption?

That's a good point, for JS types it should be completely safe (just surprising), but for Rust types like structs or `String` then it will be quite bad since it will be trying to read from a JS `undefined`.

> Basically what I'm getting at is that I believe with try_catch it doesn't actually interrupt control flow like you'd expect on an exception, right? It looks like it allows continuing execution in wasm as usual, despite the exception happening.

That's correct. The problem is that we *can't* halt execution, because `catch_unwind` doesn't work on Wasm, and there are only two ways to halt execution: Rust panics and JS exceptions.

And since `catch_unwind` doesn't work, and since JS exceptions cause stack leaks, we can't use either of those.

So I'll need to think about how to safely handle things like structs and `String`.

alexcrichton:
I do have some satellite concerns, but the main one is that an exception happening doesn't actually interrupt control flow, instead Rust sort of just continues as usual. That's the source of the runtime corruption here but even if we fixed the runtime corruption this feels too surprising to me to make feasible.

Some other concerns I would have are:

* I think the code size impact of this would want to be investigated. I suspect that a wasm module which calls one import would be significantly larger than it is today.
* You say that "it will always reset the exception handler even if f() panics" but I don't understand how that's possible? If a Rust panics happens that hits an `unreachable` instruction which isn't something that's caught here, so no destructor gets run and nothing gets reset.
* This looks like it changes the type of exception thrown if it's not caught by Rust. If I throw some sort of type error then it gets swallowed by the Rust module and reported later as a wasm `unreachable` instruction.

I definitely think that where we are today with wasm-bindgen isn't a great place. You can leak stack space and it's very difficult to work with exceptions correctly. That being said I'm not really sure how else to fix this and it sort of feels like we're in "least bad" valley right now to me.

Pauan:
> That's the source of the runtime corruption here but even if we fixed the runtime corruption this feels too surprising to me to make feasible.

We can fix it by using `Default::default()` to generate a safe value, but I agree that it's too surprising.

> You say that "it will always reset the exception handler even if f() panics" but I don't understand how that's possible?

Right, since Wasm doesn't unwind right now. But once unwinding is supported then it will work. The important thing is that it translates a JS exception into a Rust panic, and it does so in a way that doesn't leak stack/heap.

> This looks like it changes the type of exception thrown if it's not caught by Rust. 

Well that's because it's translating a JS error into a Rust panic. The same thing would happen if you `unwrap()` a `Result<T, JsValue>`.

> That being said I'm not really sure how else to fix this and it sort of feels like we're in "least bad" valley right now to me.

Yeah, since this strategy requires `Default::default()` it's really not great. I also thought about some other obscure ideas, like traits or continuation passing style, but I couldn't figure out anything good.

So unfortunately I don't think we can improve the status quo until after Wasm exceptions are available.

ctaggart:
Does it look like [Exception Handling](https://github.com/WebAssembly/exception-handling/blob/master/proposals/Exceptions.md) proposal currently has what is needed to make this work? Is there anything they are missing that would make interop easier?

Pauan:
@ctaggart All of this is really controlled by LLVM, so it depends on how LLVM handles it.

I don't think there's been any discussion yet about how LLVM will handle it, but I think `br_on_exn` should be enough.

----

@alexcrichton Incidentally, right now `panic=abort` with Wasm. Once unwinding is supported we'll be able to fix this, since unwinding will run destructors until it reaches the `try_catch`.

But even in that hypothetical future, what if somebody sets `panic=abort` in their `Cargo.toml`? How will we support that? Or will we just give up and say that `try_catch` isn't supported in that case (which is what `catch_unwind` does)?

alexcrichton:
@Pauan the loose future timeline I've envisioned is along the lines of:

* Eventually we implement `panic=unwind` for wasm. At this point users would still opt-in, since it'd by default be `panic=abort`.
* Then eventually once we're happy and everything is stable we switch to `panic=unwind` by default (to match other targets). Wasm projects can still opt-in to `panic=abort` of course.
* For `wasm-bindgen` we'll probably remove all `catch` infrastructure and move to `try_catch` like you've proposed here. This'll work (as you predict) with `panic=unwind` just fine, and I think we'll just document that with `panic=abort` nothing is ever caught, the `try_catch` would simply never catch like panicking and using `catch_unwind` today.

That's sort of my rough thinking at least, I'm not sure whether it's actually practically feasible to support `panic=abort` without a `catch` attribute.

Pauan:
@alexcrichton That all sounds reasonable to me. I also can't think of any way to support catching with `panic=abort`, at least not without some extra support from Rust itself (such as new control flow primitives).

